PR #72, opened 2014-12-08
Completely handle namedtuple pickling

### LINKED ISSUE #71: pickling nested namedtuples
While dill pickles nested classes without problem, the same cannot be said of namedtuples defined within another class:

```
In [1]: import dill

In [2]: from collections import namedtuple

In [3]: class T:
    class A: pass
    B = namedtuple("B", "")
   ...: 

In [4]: dill.dumps(T.A)
Out[4]: <elided>

In [5]: dill.dumps(T.B)
<oops!>
```

I would be OK if fixing this required manually setting some attribute on `B` to indicate where it lives, if that could help.

matsjoyce:
Fixes #71. This handles `namedtuples`, instead of handing them over to pickle, with the usual results.

Things to consider:
- In python 2.5, where there are no `namedtuple`s, unpicking will fail (not pickling). Instead of raising, should it just return `tuple`?

### DIFF
--- a/dill/dill.py
+++ b/dill/dill.py
@@ -717,6 +717,18 @@ def _create_array(f, args, state, npdict=None):
         array.__dict__.update(npdict)
     return array
 
+def _create_namedtuple(name, fieldnames, modulename):
+    mod = _import_module(modulename, safe=True)
+    if mod is not None:
+        try:
+            return getattr(mod, name)
+        except:
+            pass
+    import collections
+    t = collections.namedtuple(name, fieldnames)
+    t.__module__ = modulename
+    return t
+
 def _getattr(objclass, name, repr_str):
     # hack to grab the reference directly
     try: #XXX: works only for __builtin__ ?
@@ -1087,7 +1099,7 @@ def _proxy_helper(obj): # a dead proxy returns a reference to None
         return id(None)
     if _str == _repr: return id(obj) # it's a repr
     try: # either way, it's a proxy from here
-        address = int(_str.rstrip('>').split(' at ')[-1], base=16)   
+        address = int(_str.rstrip('>').split(' at ')[-1], base=16)
     except ValueError: # special case: proxy of a 'type'
         if not IS_PYPY:
             address = int(_repr.rstrip('>').split(' at ')[-1], base=16)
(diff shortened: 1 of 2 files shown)


mmckerns:
There is also an approach like this quick hack… it should be general, but it's messy… and as is, it breaks the namespace.

```
>>> import m
>>> import dill
>>> x = dill.dumps(compile(dill.source.importable(m.A.B, source=True), '<string>', 'exec'))
>>> exec(dill.loads(x))
>>> B
<class __main__.B at 0x10860c9a8>
```

mmckerns:
Note (as a follow-up from issue #132):
```
>>> import collections
>>> X = collections.namedtuple('Y', ['a','b'])
>>> Xi = X(0,1)
>>> import dill
>>> dill.detect.errors(Xi)
PicklingError("Can't pickle <class '__main__.Y'>: it's not found as __main__.Y",)
```
Should this be the expected behavior?  As demonstrated in `test_classdef.py`, one could conceivably pickle a badly named `namedtuple` by messing with the `__name__` (or `__qualname__`).  However, one shouldn't need to do that.  I believe that the PR for this issue will enable pickling of a badly named `namedtuple`... and is a good solution.

matsjoyce:
Do you want me to rebase this?

mmckerns:
Yes, if you don't mind.

matsjoyce:
OK, updated. I've also changed the behaviour so that it will serialize assuming that the name is bad but on deserialize it will try to find the type, and if it fails recreate it.
```py3
>>> import collections, dill
>>> X=collections.namedtuple("Y", ("a", "b"))
>>> Z=collections.namedtuple("Z", ("a", "b"))
>>> x=X(1,2)
>>> z=Z(1,2)
>>> x2=dill.copy(x)
>>> z2=dill.copy(z)
>>> type(x2)
<class '__main__.Y'>
>>> type(z2)
<class '__main__.Z'>
>>> type(x2) is X
False
>>> type(z2) is Z
True
```

matsjoyce:
Ooo, CI works. I'm looking at the failure.

matsjoyce:
Its nice having the CI, we can test aginst all the supported versions without needing to install them locally. I've fixed the failing test, so this is now merge-worthy again.

mmckerns:
I don't like that you get the same object back for a "well-formed" `namedtuple`, and you don't for a "badly-named" `namedtuple`.  That could have some unexpected consequences.  I guess we will see...

matsjoyce:
Well, we could change it so that you never got the same object back, but that would break the current behaviour. I guess we could add some gc magic at a later date?

mmckerns:
@matsjoyce: it just feels like a bit of a "wart", and those are the kind of things that usually come back to haunt you sometime later.  Now we might have objects that use a `namedtuple` returning the original object when the `namedtuple` is well-formed, and returning a copy when the `namedtuple` is not.  That makes me a bit uneasy about the PR, but like I said, I guess we will see if it's a problem.

Issue #50, opened 2014-06-26
Unpicklable objects defined in doctests interfere with serialization of lambdas


JoshRosen:
I found another example where Dill fails to pickle objects when running under doctest.  This is a one-line addition to my test-case in #18.

I'm using Python 2.7.6 and Dill eb122e6eb00dcfa9704ffa692fc7a14850867185. 

/cc @distobj

```python
import dill as pickle
import doctest

pickle.dill._trace(1)

class SomeUnreferencedUnpicklableClass(object):
    def __reduce__(self):
        raise Exception

unpicklable = SomeUnreferencedUnpicklableClass()

# This works fine outside of Doctest:
serialized = pickle.dumps(lambda x: x)

# This fails because it tries to pickle the unpicklable object:
def tests():
    """
    >>> unpicklable = SomeUnreferencedUnpicklableClass()  # <-- Added since #18
    >>> serialized = pickle.dumps(lambda x: x)
    """
    return

print "\n\nRunning Doctest:"
doctest.testmod()
```

Output:

```
F1: <function <lambda> at 0x101f8f848>
F2: <function _create_function at 0x101c96d70>
Co: <code object <lambda> at 0x100474ab0, file "dillbugtwo.py", line 13>
F2: <function _unmarshal at 0x101c96c08>
D1: <dict object at 0x10031ba20>
D2: <dict object at 0x101da9de0>


Running Doctest:
F1: <function <lambda> at 0x101f8f938>
F2: <function _create_function at 0x101c96d70>
Co: <code object <lambda> at 0x101f369b0, file "<doctest __main__.tests[1]>", line 1>
F2: <function _unmarshal at 0x101c96c08>
D2: <dict object at 0x102206120>
F1: <function tests at 0x101f8f848>
Co: <code object tests at 0x100474e30, file "dillbugtwo.py", line 16>
D1: <dict object at 0x10031ba20>
D2: <dict object at 0x10222e940>
**********************************************************************
File "dillbugtwo.py", line 19, in __main__.tests
Failed example:
    serialized = pickle.dumps(lambda x: x)
Exception raised:
    Traceback (most recent call last):
      File "/Users/joshrosen/anaconda/lib/python2.7/doctest.py", line 1289, in __run
        compileflags, 1) in test.globs
      File "<doctest __main__.tests[1]>", line 1, in <module>
        serialized = pickle.dumps(lambda x: x)
[...]

PR #86, opened 2015-03-14
serialize only the global variables used by function

davies:
This PR try to fix the issue in #50

### DIFF
--- a/dill/detect.py
+++ b/dill/detect.py
@@ -9,11 +9,18 @@ Methods for detecting objects leading to pickling failures.
 """
 
 from __future__ import absolute_import
+import sys
+import dis
 from inspect import ismethod, isfunction, istraceback, isframe, iscode
 from .pointers import parent, reference, at, parents, children
 
 from .dill import _trace as trace
 from .dill import PY3
+
+if PY3:
+    from io import StringIO
+else:
+    from StringIO import StringIO
 
 def outermost(func): # is analogous to getsource(func,enclosing=True)
     """get outermost enclosing object (i.e. the outer function in a closure)
@@ -139,6 +146,26 @@ def freevars(func):
         return {}
     return dict((name,c.cell_contents) for (name,c) in zip(func,closures))
 
+def capture_output(f):
+    stdout = sys.stdout
+    sys.stdout = StringIO()
+    f()
+    out = sys.stdout.getvalue()
+    sys.stdout = stdout
+    return out
+
+def find_globals(code):
+    out = capture_output(lambda: dis.dis(code))
+    names = set()
+    for line in out.split('\n'):
+        if '_GLOBAL' in line:
+            name = line.split('(')[-1].split(')')[0]
+            names.add(name)
+    for co in code.co_consts:
+        if co and iscode(co):
+            names |= find_globals(co)
+    return names
+
 def globalvars(func):
     """get objects defined in global scope that are referred to by func
 
@@ -154,7 +181,7 @@ def globalvars(func):
     if ismethod(func): func = getattr(func, im_func)
     if isfunction(func):
         globs = getattr(func, func_globals) or {}
-        func = getattr(func, func_code).co_names # get names
+        func = find_globals(getattr(func, func_code))
     else:
         return {}
     #NOTE: if name not in func_globals, then we skip it...
--- a/dill/dill.py
+++ b/dill/dill.py
@@ -618,16 +618,18 @@ def save_code(pickler, obj):
 
 @register(FunctionType)
 def save_function(pickler, obj):
+    from .detect import globalvars
     if not _locate_function(obj): #, pickler._session):
         log.info("F1: %s" % obj)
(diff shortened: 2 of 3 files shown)


dill/detect.py
```diff
@@ -139,6 +146,26 @@ def freevars(func):
         return {}
     return dict((name,c.cell_contents) for (name,c) in zip(func,closures))
 
+def capture_output(f):
+    stdout = sys.stdout
+    sys.stdout = StringIO()
+    f()
```
davies: This is ugly, should we parse the byte code by ourselves?

davies:
cc @joshrosen

mmckerns:
@davies: Ok, interesting… If I get the time to play with it today, I will try to get back to you this evening.  From what I see, I expect there will be a few small cleaning/changes before it goes in.

davies:
@mmckerns Thanks, any comments are welcome. detect.py depends on dill.py, so I'm not detect.py is the perfect place to put the change in.

matsjoyce:
In python 3 you could use `dis.get_instructions`, but, for compatability with 2, you could use (modified from `dis.py`):
```py
import dis

def to_int(x):
    if not isinstance(x, int):
        return ord(x)
    return x

def find_globals(co):
    globs = set()
    code = co.co_code
    n = len(code)
    i = 0
    extended_arg = 0
    while i < n:
        op = to_int(code[i])
        i = i+1
        if op >= dis.HAVE_ARGUMENT:
            oparg = to_int(code[i]) + to_int(code[i+1]) * 256 + extended_arg
            extended_arg = 0
            i = i+2
            if op == dis.EXTENDED_ARG:
                extended_arg = oparg * 65536
            if op in dis.hasname and dis.opname[op].endswith("_GLOBAL"):
                globs.add(co.co_names[oparg])
    return globs
```

mmckerns:
@davies: yeah, that's part of what should be resolved, along with some naming, and as @matsjoyce points out, some comparability issues.

davies:
@matsjoyce @mmckerns Could you take over this patch? I think that will be easier to merge.

mmckerns:
@davies: Ok, sure, I/we will do that.  I'm a bit backed up at the moment, but will get to it hopefully soon.  One thing you can do in the meantime is to make sure your patch works for test cases in #50, provide any further cases it does/doesn't work, and any further feedback on the proposed changes/requirements from #50.  Thanks for the initial submission. I'll merge it when changes/updates ready are to go.

mmckerns:
@matsjoyce: I don't get the same results from your code as I do from @davies.  Just FYI.

@davies:  I have a question for you… what would you expect from something like this:
```
a = 1
def squared(x):
  return a+x**2

def foo(x):
  def bar(y):
    return squared(x)+y
  return bar

res = globalvars(foo, shallow=False)
```

Would you expect:  ```assert set(res) == set(['squared', 'a'])``` or ```assert set(res) == set(['squared'])```?  Your code finds the latter, but it's easy to edit the code (see below) to find the former.  Do you want the former or the latter?

```
            func = find_globals(getattr(func, func_code)) #, depth=None)
            # find globals for all entries of func
            for key in func.copy():
                func.update(globalvars(globs.get(key), shallow=False).keys())
```

I definitely think what is currently in the `dill` trunk (i.e. find all global variables that are referenced in the local scope) is a desired and consistent use case.  I think the use case of finding **all globals** that are required to serialize a function is another -- and is what you were shooting at.  They should probably be separate `dill` functions.

mmckerns:
Ok, obviously this is a high-priority to wrap-up.  I'll finish this off first opportunity.

mmckerns:
edited this PR heavily, then added the 'essence' of it in: fadbffc2b688aa99d3ba29015c4cdbef463bf565

Issues arose that removed functionality from dill in certain corner cases (e.g. `test_classdef` among other tests failed), so I opted for a serialization flag on the `Pickler`.  You get the behavior you desired with `recurse=True` and it returns to the default with `recurse=False`.

If I can remove the conflict, I can remove the flag (or make the default 'recurse=True`).
Closing this PR, and will open an issue for failures with the `recurse` flag.

mmckerns:
NOTE: https://github.com/uqfoundation/dill/commit/fadbffc2b688aa99d3ba29015c4cdbef463bf565#diff-a86ebc7de6a15081b51e5b20ae78ae22R147

mmckerns:
added global setting `dill.settings['recurse']`

matsjoyce:
Looks good. Just one question: why is it called `recurse` for `dump`?

mmckerns:
The major difference is whether or not `dump` does a "shallow" pickling of `globals` or a "deep" pickling of `globals`.  This flag is also relevant for `dill.source.globalvars` and other items in `dill.source`, where the choice comes down to whether one wants to do a "shallow" pass at getting the `globals` or a "deep" one (using recursion). I started with `shallow=False` to have the code recursively find the referred-to `globals`, then didn't like the name, so opted for `recurse=True`.

PR #99, opened 2015-05-26
Check a few more prefixes for the builtin_mod test

### LINKED ISSUE #93: Pickling Error: can't pickle <class 'PyCapsule'>: it's not found as builtins.PyCapsule
I have this problem with pickling data with dill. The problem only occurs on python 3.4. My code works with python 3.3 and python 2.7.

matsjoyce:
This should fix #93, as this consults all the prefixes in the docs, plus `real_prefix`, which is from virtualenv. Could @itdxer test this? And should I be worried that this is my fifth commit on this small bit of code?

### DIFF
--- a/dill/dill.py
+++ b/dill/dill.py
@@ -944,12 +944,15 @@ def save_module(pickler, obj):
     else:
         # if a module file name starts with prefx, it should be a builtin
         # module, so should be pickled as a reference
-        base_prefix = getattr(sys, "base_prefix", sys.prefix)
-        real_prefix = getattr(sys, "real_prefix", base_prefix)
-        std_mod = (getattr(obj, "__file__", base_prefix).startswith(base_prefix)
-                   or getattr(obj, "__file__", real_prefix).startswith(real_prefix))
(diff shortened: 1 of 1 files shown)


mmckerns:
@matsjoyce: worried… probably, yes.  It's usually a sign of patching too late in the workflow as opposed to stemming it at the right spot.  However, sometimes doing so is a much harder fix. I'll wait and see if we get confirmation from @itdxer.

itdxer:
Thanks, this patch works fine for me!

mmckerns:
@matsjoyce:  I'd pull this, but it's not up-to-date with the current version of `dill.dill` in master.  I can manually merge in the changes, or you can update.  It's up to you.  Once I have a merged file, I'll check for test failure.  Then it's in.

mmckerns:
A quick manual merge of the changes you made confirms all tests pass.  If you update `dill.py` in your PR, I'll merge it right in.

matsjoyce:
Whoops, that's what I get for blindly stashing to a branch I didn't update. Sorry, fixed.

Issue #78, opened 2015-01-26
dump_session in %pylab mode


nikohansen:
In IPython: 

    import dill
    dill.dump_session('test.pkl')
    %pylab
    dill.dump_session('test.pkl')

The second dump fails with `PicklingError: Can't pickle <class 'matplotlib.axes._subplots.AxesSubplot'>: it's not found as matplotlib.axes._subplots.AxesSubplot`

IPython 2.3.1 and Python 2.7.9 from Anaconda 2.1.0 (64bit) on OSX 10.9. I can post the entire trace if desired.

nikohansen:
`%pylab` does the following:

    import numpy
    import matplotlib
    from matplotlib import pylab, mlab, pyplot 
    np = numpy
    plt = pyplot

    from IPython.core.pylabtools import figsize, getfigs

    from pylab import *
    from numpy import *

It turns out that the above problem comes from the `pylab import *`. Also `from numpy import *` leads to the pickle error `Can't pickle <type 'numpy.int64'>: it's not the same object as numpy.int64`. There is no problem without the last two `from ... import *` lines.

mmckerns:
This seems very much related to issue #4, which was closed as the `matplotlib` developers made a many of the difficult objects that `pylab` imports picklable.

The `numpy.int64` sounds like an issue that would probably be best handled on this list as a new issue, with a possible include of the `numpy` devs.  Pickling the objects coming from `pylab` are most likely to be handled by the `matplotlib` devs… however, it's quite possible that the response might be **never** import using `from pylab import *`.  If that's the case, then it may be worthwhile to include the `IPython` devs in the discussion.  I'm including @pelson, @fperez, and @charris here.

I can reproduce your errors for both, and neither are specific to `%pylab` in `IPython`.  It's a failure of the `*` import in `numpy` and `pylab`.  All of the objects just don't pickle as imported, so it's only an `IPython` concern in as much as they support a `pylab` mode.

I'm going to leave this ticket open as compatibility with `IPython` in `pylab` mode, and open a new ticket for the `numpy.int64` (and more generally `numpy import *`).  This ticket will continue for the `from pylab import *` part of the issue.

Would you mind reading  #4?  The next move may be to either submit a `matplotlib` ticket or an `IPython` ticket, depending on the response you get here.  Sound good?

mmckerns:
The namespace magic going on here is most likely the issue...

[...]

Issue #100, opened 2015-05-27
numpy array instance referenced in python data structure causes serialization of the latter to fail


lebedov:
Using dill 0.2.2 with numpy 1.9.2 and Python 2.7.9, the following runs without any problems:

```
import dill
import numpy
x = numpy.ones((2,2))
data = {}
data['x'] = x
dill.dumps(data)
```

However, if I try serializing `x` immediately before `data`, Python raises an AssertionError when attempting the second serialization:

```
import dill
import numpy
x = numpy.ones((2,2))
data = {}
data['x'] = x
dill.dumps(x)
dill.dumps(data)
```

The same problem occurs if `data` is a list and I append `x` to it.

The problem only seems to occur when `x` is a numpy array; I wasn't able to reproduce the error when I set `x` to a variety of other data types (e.g., lists, dicts, Pandas DataFrame instances, ...). Any idea why this is occurring?

matsjoyce:
Its python 2.7 only, by the looks of it:
```py
>>> import dill
>>> import numpy
>>> x = numpy.ones((2,2))
>>> data = {}
>>> data['x'] = x
>>> dill.dumps(x)
Nu: ((2, 2), float64)
F2: <function _create_array at 0x7ffb6e0b7ed8>
B2: <built-in function _reconstruct>
T4: <type 'numpy.ndarray'>
T4: <type 'numpy.dtype'>
'\x80\x02cdill.dill\n_create_array\nq\x00(cnumpy.core.multiarray\n_reconstruct\nq\x01cnumpy\nndarray\nq\x02K\x00\x85q\x03U\x01bq\x04\x87q\x05(K\x01K\x02K\x02\x86q\x06cnumpy\ndtype\nq\x07U\x02f8q\x08K\x00K\x01\x87q\tRq\n(K\x03U\x01<q\x0bNNNJ\xff\xff\xff\xffJ\xff\xff\xff\xffK\x00tq\x0cb\x89U \x00\x00\x00\x00\x00\x00\xf0?\x00\x00\x00\x00\x00\x00\xf0?\x00\x00\x00\x00\x00\x00\xf0?\x00\x00\x00\x00\x00\x00\xf0?q\rtq\x0eNtq\x0fRq\x10.'
>>> dill.dumps(data)
D2: <dict object at 0x7ffb6e053c58>
Nu: ((2, 2), float64)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
[...]

Issue #66, opened 2014-10-05
should dump_session accept a list/dict of objects to ignore?


mmckerns:
I'm not sure what kind of impact that might have if one would ignore an object... then expect to start up a session again and everything work.  Maybe it's not up to `dill` to care… and it's the user's problem if it blows things up in the dump/load of the session.

mmckerns:
Currently when I do a `dump_session`, I do something like this:
```
map(globals().pop, tuple(i for (i,j) in globals().iteritems() if not dill.pickles(j) and i not in ('__builtins__',)))
```
to remove all objects that will not pickle.  There's probably a better way to do it, but I tend to at least do some variant of the above on-the-fly.  This will not be the most efficient, but will work as long as `dill.pickles` is correct (which is overwhelmingly *most* of the time).

RuneScape314159265:
Firstly, what an epic tool! Super useful when working with jupyter notebooks that take a long time to complete and recomputing everything is either a) impossible b) merely a massive pain - thank you for making it!

I think it would be great if this ^ above hack were incorporated into the package itself, i.e. 

`dill.dump_session(fileName, ignore=True)`
Produces
`Variables:
- var1
- var3
- var5 
could not be pickled and will not be restored. Do you wish to continue (everything else that can be will be stored)? [y/n]: `

The user types yes to continue and dill saves everything about the current environment that it can, ignoring the variables specified. 

In a large file with a lot of globals I imagine that even running the check might take a while (it does in some of my files) so there should probably be an additional flag so that the user can set whether they want dill to automatically pickle even if it can't do everything - default being yes (if the user selects no, then, if dill can't pickle everything like above, they will be prompted asking whether they want to continue). 

It would be a big quality of life improvement for me, and since I'm hardly unique I'm guessing many others. 

Going hunting online for a work around isn't easy (this post, presenting the best solution, is quite hidden away and I'm guessing many don't read through it / miss it).

leogama:
Hello, I'm working on a new feature like this for `dump_session()`. As it is not always possible or convenient to delete unpickable or large but cheap to generate objects from the namespace before saving the session, as they could be needed after it, I consider this to be a relevant feature.

Before I submit a draft PR, how do you think the API could be like? And how should `load_session()`'s behavior be in this case? Should it simply ignore the not saved variables, restore them as a dill singleton as suggested, or should it do this just for variables not defined in the namespace?

I already have a working prototype that can deal with IPython's command history variables. 👌🏼

Issue #43, opened 2014-05-28
Builtin modules do not have the __dict__ pickled


matsjoyce:
Follow up for #41, number 2:

``` pycon
>>> import dill
>>> import numpy as np
>>> np.min = np.max
>>> dill.dump_session()
>>> ^D
############ restart ############
>>> import dill
>>> dill.load_session()
>>> np.min([1,2,3,4,5])
1
```

This could be fixed by deciding to pickle a module's `__dict__` (https://github.com/uqfoundation/dill/blob/master/dill/dill.py#L763-L764) using a function like:

``` py3
def _can_pickle_module(mod, _cache={}, _recursion_protection=[None]):
    if _recursion_protection[0] is not None:
        return mod is _recursion_protection[0]
    if mod not in _cache:
        _recursion_protection[0] = mod
        try:
            dumps(mod)
        except:
            _cache[mod] = False
        else:
            _cache[mod] = True
        finally:
            _recursion_protection[0] = None
    return _cache[mod]
```

and the condition in `save_module` being:

``` py3
if _can_pickle_module(obj) or is_dill(pickler) and obj is pickler._main_module:
```

However, this approach is less efficient. This needs discussion.

mmckerns:
@matsjoyce: this, #69, #70, and other related are a bit of a mess. Do you have a plan for this?  Note also that #111 etc demonstrates there are issues with this code on windows.  I keep punting on these… and they are not showing up as a high priority for me.  But, nonetheless, they are still something that needs cleaning up.

mmckerns:
@matsjoyce: You may want to check how 1938291419dc48c3b0930774b2c18272ccc3b1a4 affects the `diff` branch.  Basically, the patch I added was to pickle the `builtins` (or `__builtin__`) module by reference regardless of the value of `settings['recurse`]`.

PR #101, opened 2015-05-28
dump_session: add split_imports

### LINKED ISSUE #79: `dump_session` fails with from numpy import *
Importing numpy in this common way leads to pickle failures.

    from numpy import *
    import dill
    dill.dump_session()

Specifically:

    pickle.PicklingError: Can't pickle <type 'numpy.uint64'>: it's not the same object as numpy.uint64

and very very probably several others.

This is a bit of a big request to add all of the `numpy` top-level objects, but it's a very common case.

abrasive:
Setting split_imports=True will preprocess the module by groveling
through its top-level objects for things that are identical to those in
the other available modules. If it finds matches, it records the source
module and object names and then discards the objects during pickling.
On load, it then recovers the objects by "import x as y" using the
stored x, y.
This fixes issues with nasty unpickleable things in the depths of
libraries, as well as making the sessions smaller.

Fixes #79 and most of #78 - my pylab sessions start up with some dynamic GObject horribles in `__main__`.

### DIFF
--- a/dill/dill.py
+++ b/dill/dill.py
@@ -222,10 +222,55 @@ def loads(str):
 ### End: Shorthands ###
 
 ### Pickle the Interpreter Session
-def dump_session(filename='/tmp/session.pkl', main_module=_main_module):
+def _module_map():
+    from collections import defaultdict
+    modmap = defaultdict(list)
+    items = 'items' if PY3 else 'iteritems'
+    for name, module in getattr(sys.modules, items)():
+        if module is None:
+            continue
+        for objname, obj in module.__dict__.items():
+            modmap[objname].append((obj, name))
+    return modmap
+
+def _find_source_module(modmap, name, obj, main_module):
+    for modobj, modname in modmap[name]:
+        if modobj is obj and modname != main_module.__name__:
+            return modname
+
+def _split_module_imports(main_module):
+    modmap = _module_map()
+    imported = []
+    original = {}
+    items = 'items' if PY3 else 'iteritems'
+    for name, obj in getattr(main_module.__dict__, items)():
+        source_module = _find_source_module(modmap, name, obj, main_module)
+        if source_module:
+            imported.append((source_module, name))
+        else:
+            original[name] = obj
+    if len(imported):
+        import types
+        newmod = types.ModuleType(main_module.__name__)
+        newmod.__dict__.update(original)
+        newmod.__dill_imported = imported
+        return newmod
+    else:
+        return original
+
+def _restore_module_imports(main_module):
+    if '__dill_imported' not in main_module.__dict__:
+        return
+    imports = main_module.__dict__.pop('__dill_imported')
+    for module, name in imports:
+        exec("from %s import %s" % (module, name), main_module.__dict__)
+
+def dump_session(filename='/tmp/session.pkl', main_module=_main_module, byref=False):
     """pickle the current state of __main__ to a file"""
     f = open(filename, 'wb')
     try:
+        if byref:
+            main_module = _split_module_imports(main_module)
         pickler = Pickler(f, 2)
         pickler._main_module = main_module
         _byref = pickler._byref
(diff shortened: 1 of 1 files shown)


abrasive:
Example:

```
ipython --pylab
...
In [2]: import dill
In [3]: foo = arange(10)
In [4]: del GPollableInputStream
In [5]: del GPollableOutputStream
In [6]: del GFileDescriptorBased
In [7]: del GInitiallyUnowned
In [8]: dill.dump_session('pylab.pkl', split_imports=True)
```

Loading - no surprises here:
```
In [1]: import dill
In [2]: dill.load_session('pylab.pkl')
In [3]: foo
Out[3]: array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9])
```

abrasive:
This needs a better name than 'split_imports', I think, given that it ends up discarding most or all of the state of non-main modules. Works very well for the way I use IPython, but it's a substantial change in behaviour.

mmckerns:
@abrasive: Hmm.. thanks.  This looks nice.

@matsjoyce: We broached this question in #66, but did not implement anything as a follow-up.
Also, how does this compare to #47 and related? (I know, it's been a year.)

matsjoyce:
Well, the `use_diff` bit should be able to handle this, but it may need tweaking and more testing (I may have time in a few weeks). I did a few quick tests, and it seems to mostly work, but something in the way ipython changes objects ends up with a few unpickleable objects landing. If I understand this patch correctly, its an extension of `byref=True`, so may fit in there quite well.

mmckerns:
I'm targeting a new release for end of June, and would like to get this and #86 and #100 sorted for it. I'll go about tagging issues soon, but if you don't think this would interfere with `use_diff`, I'm likely to pull it in after some interface manipulation.  I'm starting to think kwds on `dump` (like `byref`) should be replaced with state variables on the pickler.

mmckerns:
@abrasive: I've reviewed this, and will accept it.

Could you make these mods?
 1. rename the 'split_imports` flag to `byref`, and set the default to True
 2. make it 3.x compatible (e.g. `iteritems` doesn't exist in `3.x`, etc)

I will likely rename the four helper functions, but they are fine as is for the moment.

mmckerns:
@matsjoyce: I might consider enabling this in general, not just for `dump_session`.  The only real issue I see here is that this code might lose edits to objects (e.g. `numpy.x = 1`, which appear to be neglected as is).

mmckerns:
@matsjoyce: I think this can work semi-independely of #43 (and thus #47).  It doesn't seem to remove any functionality as far as I can tell.  Nothing I have tried fails with it and not without it, but if you can think of a failure case, let me know.

mmckerns:
I take that back...  It does appear it would remove some functionality, in roughly how I expected.  If you import a pure python module and change it's `__dict__`, then it will lose those changes.  The current version of `dump_session`preserves changes (except for modules as noted in #43).

abrasive:
Amended as requested. I think having the parameter named 'byref' makes sense logically, but dump_session also deals with '_byref' which is confusing. I'll leave that up to you.

mmckerns:
@abrasive: Thanks, and true… but `byref` and `_byref` will be much more similar if if in the normal `dump`, we give this same type of behavior to modules.  Currently, the contents of pure python modules are imported upon pickling, and there should be an option to just take the import string (i.e. `byref`).  The only downside would be that you couldn't pickle a module with `byref=True` and class with `byref=False` inside of a `dump_session`.

mmckerns:
@abrasive: I was thinking something more like this, as it would allow iterators to be used for `2.x` and `3.x`:
```
def _module_map():
    items = 'items' if PY3 else 'iteritems'
    from collections import defaultdict
    modmap = defaultdict(list)
    for name, module in getattr(sys.modules, items)():
        if module is None:
            continue
        for objname, obj in getattr(module.__dict__, items)():
            modmap[objname].append((obj, name))
    return mod map
```

Would you mind?  Then I can just pull and not have to edit it.

abrasive:
OK, done.

mmckerns:
brilliant.

PR #103, opened 2015-05-31
Add IPython-specific session dump and load capability

abrasive:
IPython imports a bunch of stuff into the user namespace in Pylab and Matplotlib modes (presumably amongst others). These implicit imports are hidden from the user in commands like `%who`.

This patch is in two parts. The first modifies `dump_session`, adding a filtering capability to the new byref mechanism. A flag is also added to silently discard any byref object which cannot be found in imported modules.

The second part adds IPython dump and load functions. The dumper marks any name in the IPython hidden namespace for byref pickling. This includes the implicit imports, as well as the hidden history variables. The history variables and any dynamically generated classes end up discarded by the byref mechanism. The hidden namespace list is saved in the dumped session.
The loader loads the session in the usual way, and then fixes up the hidden namespace again according to the information saved in the dump. This ensures that a subsequent dump and load will continue to work, as well as ensuring consistent behaviour of `%who` and friends.

### DIFF
--- a/dill/__init__.py
+++ b/dill/__init__.py
@@ -24,6 +24,7 @@ __license__ = """
 """ + __license__
 
 from .dill import dump, dumps, load, loads, dump_session, load_session, \
+    dump_ipython, load_ipython, \
     Pickler, Unpickler, register, copy, pickle, pickles, \
     HIGHEST_PROTOCOL, DEFAULT_PROTOCOL, PicklingError, UnpicklingError, \
     HANDLE_FMODE, CONTENTS_FMODE, FILE_FMODE
--- a/dill/dill.py
+++ b/dill/dill.py
@@ -238,17 +238,26 @@ def _find_source_module(modmap, name, obj, main_module):
         if modobj is obj and modname != main_module.__name__:
             return modname
 
-def _split_module_imports(main_module):
+def _split_module_imports(main_module, byref_filter, discard_missing):
     modmap = _module_map()
+    if byref_filter is None:
+        byref_filter = lambda name: True
+
     imported = []
     original = {}
     items = 'items' if PY3 else 'iteritems'
     for name, obj in getattr(main_module.__dict__, items)():
+        if not byref_filter(name):
+            original[name] = obj
+            continue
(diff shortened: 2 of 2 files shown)


mmckerns:
@abrasive: I barely looked at this, but as a start: can you not assume that the `import IPython` works?  Also, consider both use and speed in `2.x` and `3.x` (I didn't look closely for specific places this is applicable).  Lastly, I don't like `dump_ipython` and `load_ipython` -- `dill` shouldn't have a separate session` dump` and `load` for `IPython` sessions, it should be integrated into the existing `dump_session` and `load_session`.

abrasive:
I didn't want to assume that all callers from IPython actually want to use IPython-specific dump and load wrappers. I'll integrate it into `dump_session` and `load_session` and remove the import when I have some time.

mmckerns:
`dump_session` has always been not quite good enough for use in`IPython`, so I appreciate the PR's.  The integration of non-standard library solutions into `dill` is always a tough call…

mmckerns:
After giving this some more thought, and a better look, I have some updated comments.  First, I don't believe `dump_ipython` and `load_ipython` don't belong in `dill`, they are probably better added to `ipython` -- since I expect this to be a point of contention, please split it off into a separate PR. As far as the other changes (filter and discard), I have no qualms with going in.  So once you rename the flags to something more concise (let's pick some names that we agree on), I'll pull the remaining changes.  Also, consider that `byref`, `filter`, and `discard` are actually *one* or *two* fields instead of *three*… if they can be condensed into less flags they should be.  A typical indication of the need for a collapse is when certain flags only make sense when another flag is set to a particular value -- this is definitely the case here, so I expect that you could collapse them.

leogama:
@mmckerns I'm working on this issue with a different kind of solution. Code is growing. What do you think of splitting "session" related code to a submodule? Beyond its size, it also lays down in the way when navigating between imports and definitions and the `Pickler` and `Unpickler` code in `_dill.py`.

mmckerns:
I would think that the `session` related code is pretty tied up with the other code, and hard to separate.  It would be interesting to see this attempted...

PR #113, opened 2015-06-26
Fix some win32  issues:

mindw:
* handle lack of curses gracefully
* CDLL(NULL) throws an exception in py27 windows.

Also assume ctype always available.

### DIFF
--- a/dill/_objects.py
+++ b/dill/_objects.py
@@ -56,7 +56,6 @@ import hmac
 import os
 import logging
 import optparse
-import curses
 #import __hello__
 import threading
 import socket
@@ -69,12 +68,8 @@ try:
     HAS_ALL = True
 except ImportError: # Ubuntu
     HAS_ALL = False
-try:
-    import ctypes
-    HAS_CTYPES = True
-except ImportError: # MacPorts
-    HAS_CTYPES = False
-from curses import textpad, panel
+
+import ctypes
 
 # helper objects
 class _class:
@@ -106,11 +101,11 @@ def _function2():
         from sys import exc_info
         e, er, tb = exc_info()
         return er, tb
-if HAS_CTYPES:
-    class _Struct(ctypes.Structure):
-        pass
-    _Struct._fields_ = [("_field", ctypes.c_int),("next", ctypes.POINTER(_Struct))]
+class _Struct(ctypes.Structure):
+    pass
+_Struct._fields_ = [("_field", ctypes.c_int),("next", ctypes.POINTER(_Struct))]
 _filedescrip, _tempfile = tempfile.mkstemp('r') # deleted in cleanup
+os.close(_filedescrip)
 _tmpf = tempfile.TemporaryFile('w')
 
 # put the objects in order, if possible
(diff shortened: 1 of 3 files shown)


mmckerns:
I agree with some of this, and am working on windows issues… so this is timely.  However, assuming `ctypes` is a bad assumption.  It is not available on all platforms, for example Google's AppEngine: https://cloud.google.com/appengine/kb/libraries.  Could you back out that change?

mindw:
done + setup.py fix.

mmckerns:
I noticed the `pyreadline` issue this morning.  Your patch does not fix it… just part of it.  Could you convert `pyreadline` to `pyreadline_version` everywhere?  The issue is actually where it's declared.  Secondly, this should read ```      print ("    pyreadline %s" % pyreadline_version)```.  And, while you are at it, for consistency, `objgraoh` should be `objgraph_version` everywhere.

dill/_objects.py
```diff
@@ -468,14 +471,23 @@ class _Struct(ctypes.Structure):
...
+    #x['CursesPanelType'] = panel.new_panel(_curwin)
+except ImportError:
+    pass
+    
 if HAS_CTYPES:
     x['CCharPType'] = ctypes.c_char_p()
     x['CWCharPType'] = ctypes.c_wchar_p()
     x['CVoidPType'] = ctypes.c_void_p()
-    x['CDLLType'] = _cdll = ctypes.CDLL(None)
+    if sys.platform == 'win32':
```
mmckerns: more robust is `sys.platform[:3] == 'win'`

dill/_objects.py
```diff
@@ -111,6 +109,7 @@ class _Struct(ctypes.Structure):
         pass
     _Struct._fields_ = [("_field", ctypes.c_int),("next", ctypes.POINTER(_Struct))]
 _filedescrip, _tempfile = tempfile.mkstemp('r') # deleted in cleanup
+os.close(_filedescrip)
```
mmckerns: There's probably a reason I didn't do that.  I'm assuming that it changes the file descriptor from an open one to a closed one… and those pickle differently.  I'll have to take a closer look, but please punt this to a new PR, and I can merge the rest of this.

dill/_objects.py
mindw: This follows the environment markers syntax:
https://www.python.org/dev/peps/pep-0426/#environment-markers
It stans for all window platforms where ctypes was built with Visual Studio AFAIK.

dill/_objects.py
mindw: It has to be closed on windows or the _tempfile  deletion will fail .

dill/_objects.py
mmckerns: That PEP is too recent as per the platforms and versions supported, so I have to use the more robust version.

dill/_objects.py
mmckerns: Yes, I noticed that list night as I was debugging on windows. Therefore it needs to be handled as is for non-windows systems, and changed for windows systems.  This might also impact the tests below.  So… new PR.  It's a bit more to investigate than a simple pull.

dill/_objects.py
mindw: This was just an example. The previous one https://www.python.org/dev/peps/pep-0345/#environment-markers has the same syntax. This the syntax used and implemented by pip and setuptools.

mmckerns: I'm aware it is, and has been the standard for windows 32 and 64 bit systems for a long time.  `dill` is getting on to be 10 years old, so it could be an artifact.  However, I do remember having issues with `win*` versus `win32`, but can't point to anything recent offhand.

dill/_objects.py
mindw: When running test_objects.py both cases gave the same results
1. close file descriptor & delete tempfile
2. do neither

result:

```bat
fails: CLibraryLoaderType <class 'ctypes.LibraryLoader'>
fails: TextWrapperType <type 'file'>
fails: BufferedRandomType <type 'file'>
fails: BufferedReaderType <type 'file'>
fails: PyTextWrapperType <class '_pyio.TextIOWrapper'>
fails: PyBufferedRandomType <class '_pyio.BufferedRandom'>
fails: PyBufferedReaderType <class '_pyio.BufferedReader'>
fails: TarFileType <class 'tarfile.TarFile'>
fails: FileType <type 'file'>
fails: ListIteratorType <type 'listiterator'>
fails: TupleIteratorType <type 'tupleiterator'>
fails: XRangeIteratorType <type 'rangeiterator'>
fails: CycleType <type 'itertools.cycle'>
fails: TemporaryFileType <type 'instance'>
fails: PermutationsType <type 'itertools.permutations'>
fails: CombinationsType <type 'itertools.combinations'>
fails: RepeatType <type 'itertools.repeat'>
fails: CompressType <type 'itertools.compress'>
```

mmckerns: That's because that line is not tested.  To check it, comment the former and uncomment the latter.
```
load_types(pickleable=True,unpickleable=False)
#load_types(pickleable=True,unpickleable=True)
```

mmckerns:
I appreciate the work on a small patch.  I'll take it after the last change noted above.

PR #118, opened 2015-07-30
Handle cases where os.path.exists throws

### LINKED ISSUE #117: TypeError for TemporaryFile in _create_filehandle
@matsjoyce: looks like name of temporary file is passed as an int, and is expected to be a string.
```
Python 3.2.6 (default, Oct 15 2014, 21:18:54) 
[GCC 4.2.1 Compatible Apple Clang 4.1 ((tags/Apple/clang-421.11.66))] on darwin
>>> import tempfile
>>> tempfile.TemporaryFile('w')
<_io.TextIOWrapper name=3 mode='w' encoding='US-ASCII'>
>>> _tmpf = _
>>> 
>>> import dill
>>> dill.loads(dill.dumps(_tmpf))
(<class 'TypeError'>, TypeError("Can't convert 'int' object to str implicitly",), <traceback object at 0x10c8b06c8>)
<function _create_filehandle at 0x10c6f9a68> (3, 'w', 0, False, <built-in function open>, False, 0, '')
  File "/Users/mmckerns/lib/python3.2/site-packages/dill-0.2.4.dev0-py3.2.egg/dill/dill.py", line 209, in loads
    return load(file)
  File "/Users/mmckerns/lib/python3.2/site-packages/dill-0.2.4.dev0-py3.2.egg/dill/dill.py", line 199, in load
    obj = pik.load()
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.2/lib/python3.2/pickle.py", line 834, in load
    dispatch[key[0]](self)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.2/lib/python3.2/pickle.py", line 1138, in load_reduce
    value = func(*args)
  File "/Users/mmckerns/lib/python3.2/site-packages/dill-0.2.4.dev0-py3.2.egg/dill/dill.py", line 498, in _create_filehandle
    if not os.path.exists(name):
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.2/lib/python3.2/genericpath.py", line 18, in exists
    os.stat(path)
TypeError: Can't convert 'int' object to str implicitly
```

matsjoyce:
This should fix #117. It seems to give the right behaviour, but could theoretically play up in circumstances I haven't thought of.

Also move ctypes check up, as the current behaviour could cause problems.

### DIFF
--- a/dill/dill.py
+++ b/dill/dill.py
@@ -494,8 +494,11 @@ def _create_filehandle(name, mode, position, closed, open, strictio, fmode, fdat
         # treat x mode as w mode
         if "x" in mode and sys.hexversion < 0x03030000:
             raise ValueError("invalid mode: '%s'" % mode)
-
-        if not os.path.exists(name):
+        try:
+            exists = os.path.exists(name)
+        except:
+            exists = False
+        if not exists:
             if strictio:
                 raise FileNotFoundError("[Errno 2] No such file or directory: '%s'" % name)
             elif "r" in mode and fmode != FILE_FMODE:
@@ -537,6 +540,8 @@ def _create_filehandle(name, mode, position, closed, open, strictio, fmode, fdat
                     r = getattr(r, "raw", r)
                     r.name = name
                 else:
+                    if not HAS_CTYPES:
+                        raise ImportError("No module named 'ctypes'")
                     class FILE(ctypes.Structure):
                         _fields_ = [("refcount", ctypes.c_long),
                                     ("type_obj", ctypes.py_object),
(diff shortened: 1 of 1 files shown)


mmckerns:
that's what it needed

mmckerns:
you weren't up to date, were you?  I think you were missing the latest commits… and I didn't notice that

mmckerns:
Hmm… this patch broke other recent fixes… and the merge will reopen #117

matsjoyce:
Whoops! I wish git did that for me...

PR #166, opened 2016-05-13
Normalize module paths when checking prefixes

njwhite:
Dill keeps trying to serialize my `sys` module as its prefix directory isn't normalized:

```
In [11]: sys.prefix
Out[11]: '/Users/me/project/bin/.envname/bin/..'
```


matsjoyce:
Looks OK to me.

PR #162, opened 2016-03-16
Make tests/test_check.py py.test friendly

lesteve:
First go at tackling #139 one file at a time.

Adding `__init__.py` in the tests folder is a convenience so that you can do `py.test tests/test_check.py`. It seems to be a convention in all the projects I have worked with too (numpy, scipy, scikit-learn, etc ...).

Issue #139, opened 2015-11-26
How to run the tests in a convenient way?


lesteve:
Sorry if i haved missed it somewhere in the documentation, but how do you run the dill tests in a convenient way? I am used to projects where nosetests is used but trying with dill yields plenty of errors.

Here is what I did:

```
git clone https://github.com/uqfoundation/dill
cd dill
python setup.py build
nosetests
```

While I am at it, `nosetests --with-doctest` shows some failing doctests.

Would you be interested in contributions that make the tests more nosetests friendly? 

I saw that there were some test files in dill/tests that you could run directly with python but it is unclear whether a test failure is expected or not. For example, here is the output I get with master for tests/test_objects.py:

```
❯ python tests/test_objects.py
fails: ListIteratorType <type 'listiterator'>
fails: TupleIteratorType <type 'tupleiterator'>
fails: XRangeIteratorType <type 'rangeiterator'>
fails: CycleType <type 'itertools.cycle'>
fails: PermutationsType <type 'itertools.permutations'>
fails: CombinationsType <type 'itertools.combinations'>
fails: RepeatType <type 'itertools.repeat'>
fails: CompressType <type 'itertools.compress'>
```

mmckerns:
See #138.

The tests in `dill/tests` should not produce any output -- with the exception of `dill/tests/test)objects.py`.  `dill/tests/test_objects.py` prints a collection of objects that fail to serialize with `dill`.  The plan is to have it set up for `nose` (or similar) to conveniently run all tests.

lesteve:
> See #138.

I see, I somehow missed this issue completely.

I would be keen on working on making the tests runnable with nosetests if you are interested.

mmckerns:
@lesteve: It's on my to do list… so if you want to do it, go for it.

disconnect3d:
@lesteve Is there any progress on it?

lesteve:
Not any significant progress I am afraid, although I haven't given up on it completely. Last time I tried I got tripped up by the fact that some tests in test_classdef.py pass fine when run with python but not with nosetests. My understanding was that dill seems to behave differently when `__name__ == __main__`, which isn't the case when the tests are run with nosetests. I never got to the bottom of this.

lesteve:
I decided to tackle this one test file at a time. The idea is to keep the current behavior of being able to do `for i in tests/*.py; do python $i; done` while I am making the tests py.test friendly.

The main idea is to have a structure like this for each test file:

```python
[...]

PR #190, opened 2016-11-18
On Py3, use cPickle Unpickler

kernc:
Pure-Python unpickler, available as `pickle._Unpickler`, uses `struct` to unpack
data, and the latter can also fail with `struct.error`. The default
C unpickler fails with documented and expectable `UnpicklingError`, and
is also faster.

Dill's unpickler seems to be a very thin wrapper around pickle's unpickler, so the change is not grave in any way, just somewhat more convenient for the end user.

-----
I'd like to help set up continuous integration build testing on Travis. What is the canonical way to run and depend on tests in the _tests_ folder? If the files are run sequentially and all asserts pass, the build is considered passing?

### DIFF
--- a/dill/dill.py
+++ b/dill/dill.py
@@ -37,7 +37,7 @@ _use_diff = False
 PY3 = (sys.hexversion >= 0x30000f0)
 if PY3: #XXX: get types from .objtypes ?
     import builtins as __builtin__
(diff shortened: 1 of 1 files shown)


matsjoyce:
For the CI/tests part, see #3, #139, #162. Its been on the burner for quite a while... :smile:

PR #191, opened 2016-11-18
Enable Travis CI

kernc:
Contrary to other CI attempts, I choose the least invested route that is hopefully very easy to review and merge.

This just runs the test examples on Travis, reporting any failures. When you enable CI for this repo on Travis, as you can see, [the build passes](https://travis-ci.org/kernc/dill/builds/177006898), except for the newer Python versions which you will eventually (hopefully) bring up to speed. :smile:

Python <=2.6 support is not tested. Nobody uses that any more.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,28 @@
+language: python
+
+sudo: false
+
+matrix:
+    include:
+        - python: '2.7'
+        - python: '3.4'
+        - python: '3.5'
+        - python: '3.6-dev'
+        - python: 'nightly'
+    allow_failures:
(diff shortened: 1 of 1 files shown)


kernc:
@mmckerns How does this look to you? :smiley:  One need only [enable the repo on Travis](https://travis-ci.org/profile/uqfoundation) to see it working quite well.

mmckerns:
@kernc: This looks good.  `dill` needs to be using CI. I wanted to be using it already a few years ago, and just have not gotten around to it.  So I will review this and #198, and see if we can PR one or both.

Issue #175, opened 2016-06-30
dill_bugs breaks in pypy on save_wrapper_descriptor


mmckerns:
As reported by @peque:

```
$ pypy tests/dill_bugs.py 
True: <type 'str'>, __main__
True: <type 'function'>, <function _method at 0x0000000002534ea8>
Traceback (most recent call last):
  File "tests/dill_bugs.py", line 42, in <module>
    print ("%s: %s, %s" % (dill.pickles(i), type(i), i))
  File "/media/src/dill/dill/dill.py", line 1304, in pickles
    pik = copy(obj, **kwds)
  File "/media/src/dill/dill/dill.py", line 206, in copy
    return loads(dumps(obj, *args, **kwds))
  File "/media/src/dill/dill/dill.py", line 256, in dumps
    dump(obj, file, protocol, byref, fmode, recurse)#, strictio)
  File "/media/src/dill/dill/dill.py", line 249, in dump
    pik.dump(obj)
  File "/usr/lib64/pypy-4.0.1/lib-python/2.7/pickle.py", line 224, in dump
    self.save(obj)
  File "/usr/lib64/pypy-4.0.1/lib-python/2.7/pickle.py", line 286, in save
    f(self, obj) # Call unbound method with explicit self
  File "/media/src/dill/dill/dill.py", line 1008, in save_wrapper_descriptor
    pickler.save_reduce(_getattr, (obj.__objclass__, obj.__name__,
AttributeError: generic property has no __objclass__
```

Peque:
Are there any updates on this matter? :blush:

Peque:
Hi Mike, is there any update on this matter? :innocent:

mmckerns:
@Peque: if there were, I'd update this ticket.  I believe this is all that is needed for you to have the compatibility you want, but I have not had time to resolve this issue with the `wrapper_descriptor` just yet.

Peque:
@mmckerns I understand. Thanks a lot for taking the time to reply anyway! :blush:

mmckerns:
I have opened a ticket on `pypy` at: https://bitbucket.org/pypy/pypy/issues/2464/getset_descriptor-cannot-access

mmckerns:
There's now a patch for this in `pypy`.  After some testing against the latest `pypy`, I think this should be able to be closed.

mmckerns:
[...]

Issue #132, opened 2015-10-05
badly named namedtuple fails in python 3.5 and 3.6


mmckerns:
For `python3.5` one test fails in `test_classdef.py`:

```
>>> from collections import namedtuple
>>> Z = namedtuple("Z", ['a','b'])
>>> Zi = Z(0,1)
>>> X = namedtuple("Y", ['a','b'])
>>> X.__name__ = "X"
>>> Xi = X(0,1)
>>> import dill
>>> assert Z == dill.copy(Z)   
>>> assert Zi == dill.copy(Zi)
>>> dill.detect.errors(X)
PicklingError("Can't pickle <class '__main__.Y'>: it's not found as __main__.Y",)
>>> dill.detect.trace(True)
>>> assert X == dill.copy(X)
T6: <class '__main__.Y'>
Traceback (most recent call last):
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/pickle.py", line 268, in _getattribute
    obj = getattr(obj, subpath)
AttributeError: module '__main__' has no attribute 'Y'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/pickle.py", line 907, in save_global
    obj2, parent = _getattribute(module, name)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/pickle.py", line 271, in _getattribute
    .format(name, obj))
AttributeError: Can't get attribute 'Y' on <module '__main__' (<_frozen_importlib_external.SourceFileLoader object at 0x10b3910b8>)>

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/mmckerns/lib/python3.5/site-packages/dill-0.2.5.dev0-py3.5.egg/dill/dill.py", line 161, in copy
    return loads(dumps(obj, *args, **kwds))
  File "/Users/mmckerns/lib/python3.5/site-packages/dill-0.2.5.dev0-py3.5.egg/dill/dill.py", line 197, in dumps
    dump(obj, file, protocol, byref, fmode, recurse)#, strictio)
  File "/Users/mmckerns/lib/python3.5/site-packages/dill-0.2.5.dev0-py3.5.egg/dill/dill.py", line 190, in dump
    pik.dump(obj)
[...]

PR #114, opened 2015-06-26
Refactor setup.py:

mindw:
* use entry_points
* build universal wheels
* use setuptools_scm for version tracking
* eliminate readme duplication by dropping README.md
* use manifest.in
* add .gitignore

For your consideration.

PR #198, opened 2016-12-17
Pytest and CI

Peque:
This pull request implements initial `pytest` compatibility and CI with Travis. [See the results after setting up the repository with Travis](https://travis-ci.org/Peque/dill).

Tests are executed with `tox`, meaning that the user can very easily reproduce all test cases (with different Python interpreters) locally and easily. `tox` calls `pytest` to run those test files that have been adapted. Then, the rest of the files, are tested as before (simply executing the test file with the interpreter).

It is a step forward to CI and `pytest` integration. Although it is not complete, I think it could be merged as-is (I don't see any downsides).

Locally, I have already adapted most of the test files to be `pytest` compatible. However, they all fail at least at one point. Therefore, I would need more time to fix them and, probably, discuss them with you @mmckerns to understand what the test is doing and why is now failing.

Any comments/suggestions?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,4 @@
+.tox/
+.cache/
+*.egg-info/
+*.pyc
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,10 @@
+language: python
+
+python:
+  - 3.5
+
+install:
+  - pip install tox
+
+script:
+  - tox
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,2 @@
+include README
+include LICENSE
--- a/dill/__init__.py
+++ b/dill/__init__.py
@@ -24,7 +24,7 @@ __license__ = """
 """ + __license__
 
 from .dill import dump, dumps, load, loads, dump_session, load_session, \
-    Pickler, Unpickler, register, copy, pickle, pickles, check, \
+    Pickler, Unpickler, register, copy, pickle, pickles, \
     HIGHEST_PROTOCOL, DEFAULT_PROTOCOL, PicklingError, UnpicklingError, \
     HANDLE_FMODE, CONTENTS_FMODE, FILE_FMODE
 from . import source, temp, detect
--- a/dill/dill.py
+++ b/dill/dill.py
@@ -16,7 +16,7 @@ Test against CH16+ Std. Lib. ... TBD.
 """
 __all__ = ['dump','dumps','load','loads','dump_session','load_session',
            'Pickler','Unpickler','register','copy','pickle','pickles',
-           'check','HIGHEST_PROTOCOL','DEFAULT_PROTOCOL','PicklingError',
+           'HIGHEST_PROTOCOL','DEFAULT_PROTOCOL','PicklingError',
            'UnpicklingError','HANDLE_FMODE','CONTENTS_FMODE','FILE_FMODE']
 
 import logging
@@ -1075,7 +1075,7 @@ def _proxy_helper(obj): # a dead proxy returns a reference to None
         return id(None)
     if _str == _repr: return id(obj) # it's a repr
     try: # either way, it's a proxy from here
-        address = int(_str.rstrip('>').split(' at ')[-1], base=16)   
+        address = int(_str.rstrip('>').split(' at ')[-1], base=16)
     except ValueError: # special case: proxy of a 'type'
         if not IS_PYPY:
             address = int(_repr.rstrip('>').split(' at ')[-1], base=16)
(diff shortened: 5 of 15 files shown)


Peque:
Related issues/pull requests: #3, #139, #162, #191, #198

Also #114 (just for the inclusion of a `MANIFEST.in` file, which was necessary to run the tests).

matsjoyce:
Looks good :+1: . What we could do it merge this into a separate branch, iron out the bugs there first, and then merge it into master which will stop the false negatives for other pull requests while we work on this.

Peque:
@matsjoyce As it is right now it should not generate false negatives (the bugs are just with the changes I have locally in my computer), so maybe it could directly be merged to master as a step forward?

Even if it stays like this forever (thinking about the worst case, in which we can not deal with the bugs when adapting the rest of the files), I think it is better this way. What you think?

matsjoyce:
Getting CI can only be a good thing unless it generates tons of spam, which travis shouldn't. I think we should merge ASAP, but it's not my call.

mmckerns:
This looks good. As I mentioned in #191, `dill` needs to be using CI -- I wanted to be using it already a few years ago, and just have not gotten around to it. So I will review this and #191, and see if we can PR one or both.

Peque:
@mmckerns Thanks for considering this pull request.

It has few differences with respect to #191:

- It is an attempt not only to enable Travis CI, but to refactor/split some tests to make their output more useful.
- It also uses `tox` to allow the user (not only Travis) to run all tests locally with a single command and against multiple interpreters.
- As in #191, it also runs the tests that have not been refactored/split yet "the old way" (executing the files with the python interpreter). However, I think #191 will stop when the first error occurs, while this implementation continues until the end (while still reporting a fail at the end if it happened, of course).

mmckerns:
@Peque: Yes, I see the differences -- but thanks for itemizing them in a compact way.

mmckerns:
I'm going to start with the PR in #191.  I like a lot of what is in your PR, and I think a lot of it should be used in `dill`.  I'm going to test and explore a little more, then get back to you on this PR.

mmckerns:
@Peque: can you move the conflicting `travis.yml` into a separate PR, so we split this into a PR for (1) using `pytest`, and (2) hooking `pytest` into `travis` (I assume with `tox`)?  Then I think we can get the `pytest` stuff pulled pretty quickly.

Peque:
@mmckerns Thanks for taking the time to review the changes. Next time, if you can, add your comments associated to the specific changes, because that way it is easier for me to know to which code part you are referring to (I think you just clicked on the "global" <Review Changes> button for the pull request).

> don't move check from dill to test_check. Otherwise, the change to test_check looks good.

Ok, just thought it was really a test-only function, which didn't make sense as actual dill code.

> this change makes pytest required to run the tests. I'd rather not do that. Otherwise, the changes to test_file are fine.

I understand you would want, at first, to avoid that. However, `pytest` has many "nice" things that could help make other tests much nicer, so I would vote (if I may) for adding it as the official tool to run dill tests. Really many projects now are using `pytest` as the "de facto" testing tool nowadays.

> ultimately I think your tox.ini and travis.yml are a better option than PR #191, and is a fine start for adding tox and pytest support. I also see the test failures with pytest and need to better understand why. I'll come back to approving your tox.ini and travis.yml later.

If you like tox, I can easily create a pull request for that only. I think it is a good change anyway. It allows you to run all the tests using multiple interpreters locally (then we can discuss adding the `pytest` stuff in another pull request).

> Why run with pytest at all if you are going to run the tests with bash?

Bash runs only old tests (new modified tests do not run any code when executed with `python testX.py`). New tests only have functions defining each separate test. So it is a way to make sure all tests are run (when possible, using pytest, otherwise, running them with the interpreter).

> dill works for 2.6-2.7 and 3.1-3.6 and pypy and pypy3, with some minor errors (as seen in test suites). Do you mean "unsupported" by there is a single test failure, or otherwise?

I think adding `tox` support for now may help you run the tests locally. As far as I have seen (and Travis seems to agree with me), Python 3.5, 3.6 and pypy (2 and 3) cannot run the test suite without errors. Are you able to run the tests with those interpreters locally?

> @Peque: can you move the conflicting travis.yml into a separate PR, so we split this into a PR for (1) using pytest, and (2) hooking pytest into travis (I assume with tox)? Then I think we can get the pytest stuff pulled pretty quickly.

If you agree, as I have mentioned, we could start by simply integrating `tox`. Then we can start migrating old tests to `pytest` and discuss the errors that the migration may raise (I could create another pull request for that).

mmckerns:
@Peque: yeah sorry about the dissociated comments, I was trying something new and it didn't work as I expected.

Peque:
@mmckerns Notice that, when you hover your mouse on the code (on the "Files Changes" tab of the pull request), there is a `+` sign that appears and allows you to add a comment there. That may work better. :blush:

mmckerns:
> I understand you would want, at first, to avoid that. However, pytest has many "nice" things that could help make other tests much nicer, so I would vote (if I may) for adding it as the official tool to run dill tests. Really many projects now are using pytest as the "de facto" testing tool nowadays.

Yes, I know… but bear with me.  I'm going to use `pytest`, but not immediately. Maybe in a day or two, when I get all of this sorted. So, if you wouldn't mind, let's separate the PR into separate PRs.  I'd like one PR that changes the test code to make it more `pytest` friendly (without requiring `pytest`), and one PR for `tox` only is fine, and then I think what is left is the `travis.yml` file, and potentially baking in `pytest`. 

I would love to use `pytest` for `travis`… but not to force someone who wants to try the tests to install `pytest`.  I've used `pytest` before several times on other projects, and it's nice, yes.

> Bash runs only old tests (new modified tests do not run any code when executed with python testX.py). New tests only have functions defining each separate test. So it is a way to make sure all tests are run (when possible, using pytest, otherwise, running them with the interpreter).

Aha, I didn't register that you had deleted the `__main__` block.  I see, yeah. Duh.

> I think adding tox support for now may help you run the tests locally. As far as I have seen (and Travis seems to agree with me), Python 3.5, 3.6 and pypy (2 and 3) cannot run the test suite without errors. Are you able to run the tests with those interpreters locally?

The tests pass locally for 3.5 and 3.6, with the exception of one test (namedtuple) in `test_classdef`, and there's a ticket for that #132, and also pypy has only one test fail `dill_bugs` (with issue #175)

Peque:
Closed this in favor of #199 and #200.

mmckerns:
thanks

PR #200, opened 2016-12-28
Pytest compatibility

Peque:
These changes implement some `pytest` compatibility in tests. Note that, in order to keep compatibility with older test-running methods, an `if __name__ == '__main__'` section has been created in all test files to force execution when running tests directly with the interpreter.

This is really inconvenient and I am against it, as adding tests would imply the need to add explicit calls within that section. I guess it could be okay as a transition, though. And it is also impossible to avoid as long as `pytest` is not adopted as a dependency for running (adapted) dill tests. :wink:

### DIFF
--- a/tests/test_check.py
+++ b/tests/test_check.py
@@ -8,14 +8,14 @@
 
 from __future__ import with_statement
 from dill import check
-from dill.temp import capture
-from dill.dill import PY3
 import sys
 
-f = lambda x:x**2
+from dill.temp import capture
+from dill.dill import PY3
+
 
 #FIXME: this doesn't catch output... it's from the internal call
-def test(func, **kwds):
+def raise_check(func, **kwds):
     try:
         with capture('stdout') as out:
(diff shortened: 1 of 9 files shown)


mmckerns:
Can you (1) update this PR, (2) delete `test_file.py` from this PR... or don't make it explicitly depend on the import of `pytest` and make sure that the `.txt` file artifact is deleted?

Otherwise, it's ready to PR, and I'll address the conversion of the other tests myself.

mmckerns:
@Peque: Thanks, btw, for pushing on this and the other PRs.  I'm working through it -- I will also see what I can do to address the issue that you are most interested in for `pypy` (i.e. #175).

Peque:
@mmckerns Rebased. Removed `pytest` dependency and fixed `test_file.py` clean-up/teardown.

mmckerns:
Thanks, I'll take it from here.

PR #202, opened 2017-01-06
Make test_file.py tests parallelizable

Peque:
Just a small change that makes tests in `test_file.py` parallelizable. This is useful for being able to run tests in parallel with `pytest` (i.e.: `pytest -n 8`).

*Note that `pytest-xdist` module is required in order to use the `-n` option with `pytest`.*


mmckerns:
This file takes little to no time to run as is.  Your point is that with these changes, all the tests can run in parallel?

Peque:
@mmckerns That is exactly my point, yes. :blush:

matsjoyce:
Futureproofing!

Peque:
@mmckerns Removed unnecessary print.

Issue #187, opened 2016-10-25
Downgrade logging level from info to debug


rs2:
Pickling an object with the logging level of root logger set to info results in lots of unreadable text of the following form:

```
T1: <class 'classmethod'>
D2: <dict object at 0x00000000>
T1: <class 'OSError'>
D2: <dict object at 0x00000000>
```

Please consider downgrading the level of logging to `debug` in e.g. https://github.com/uqfoundation/dill/blob/master/dill/dill.py#L814

mmckerns:
I totally realize what I have implemented is a bit odd, and unexpected verbosity may occur if logging level is set to `info`... but to me, it's the correct behavior.  Those messages are not debug messages, they are "info" from the `dill.detect.trace` method.  (as I noted in your PR).

mmckerns:
I'm going to close this... but feel free to continue the discussion in the PR you've submitted.

matsjoyce:
I'm not sure how this is an issue anyway, as the `__init__` module sets the level of the dill logger to waning. @rs2 How are you setting your logging up, because unless you set dill's logger to info yourself, there shouldn't be a clash?

PR #205, opened 2017-01-22
#187 Downgrade logging level from info to debug

rs2:
As discussed in the issue, `dill` can be way is too verbose, e.g. for map-reduce applications. 

The messages emitted are more suitable for debugging `dill` rather than informing the developers using `dill`. Hence downgrading to `log.debug`.

PS Not every log.debug entry has been replaced.

Release notes entry:

#187 Emit debug messages with lower priority


mmckerns:
Doing so (as you have PR'd so far) breaks the functionality of `dill.detect.trace`, which is intended for users to be able to have a pickle trace.  You'd need to fix that.

mmckerns:
Can you provide a reason the level should be modified?   If not, I'm going to close this.

PR #214, opened 2017-03-07
use with statement instead of try...finally

stegben:
Kind of a PEP8 issue. It's easier to read.

### DIFF
--- a/dill/dill.py
+++ b/dill/dill.py
@@ -339,8 +339,7 @@ def dump_session(filename='/tmp/session.pkl', main=None, byref=False):
     from .settings import settings
     protocol = settings['protocol']
     if main is None: main = _main_module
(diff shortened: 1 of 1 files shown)


matsjoyce:
The reason it is not used is that python 2.5 may not have it, and that is a currently supported version of python for dill. If you want to use that, you need `from __future__ import with_statement`.

stegben:
Got it.

mmckerns:
regardless, thanks for submitting the PR

matsjoyce:
We can merge it if and when we drop 2.5 :grin:

mmckerns:
@matsjoyce: absolutely

stegben:
@mmckerns You're welcome. `dill` is a really useful package.

Issue #261, opened 2018-04-13
Can't use `recurse=True` with Numpy array. 


marius311:
The following code,
```python
from dill import dumps
from numpy import array
x = array([1,2,3])
f = lambda: x
dumps(f,recurse=True)
```
gives

```python-traceback
ValueError                                Traceback (most recent call last)
<ipython-input-1-1ea050a9a020> in <module>()
      3 x = array([1,2,3])
      4 f = lambda: x
----> 5 dumps(f,recurse=True)

~/.local/lib/python3.5/site-packages/dill/dill.py in dumps(obj, protocol, byref, fmode, recurse)
    279     """pickle an object to a string"""
    280     file = StringIO()
--> 281     dump(obj, file, protocol, byref, fmode, recurse)#, strictio)
    282     return file.getvalue()
    283 

~/.local/lib/python3.5/site-packages/dill/dill.py in dump(obj, file, protocol, byref, fmode, recurse)
    272         raise PicklingError(msg)
    273     else:
--> 274         pik.dump(obj)
    275     stack.clear()  # clear record of 'recursion-sensitive' pickled objects
    276     return

/usr/lib/python3.5/pickle.py in dump(self, obj)
    406         if self.proto >= 4:
    407             self.framer.start_framing()
--> 408         self.save(obj)
    409         self.write(STOP)
    410         self.framer.end_framing()

/usr/lib/python3.5/pickle.py in save(self, obj, save_persistent_id)
    473         f = self.dispatch.get(t)
    474         if f is not None:
--> 475             f(self, obj) # Call unbound method with explicit self
    476             return
    477 

~/.local/lib/python3.5/site-packages/dill/dill.py in save_function(pickler, obj)
   1329             # recurse to get all globals referred to by obj
   1330             from .detect import globalvars
-> 1331             globs = globalvars(obj, recurse=True, builtin=True)
   1332             # remove objects that have already been serialized
   1333            #stacktypes = (ClassType, TypeType, FunctionType)

~/.local/lib/python3.5/site-packages/dill/detect.py in globalvars(func, recurse, builtin)
    212             for key in func.copy(): #XXX: unnecessary...?
    213                 nested_func = globs.get(key)
--> 214                 if nested_func == orig_func:
    215                    #func.remove(key) if key in func else None
    216                     continue  #XXX: globalvars(func, False)?

ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

[...]

PR #280, opened 2018-08-13
Fix comparing algorithm for serialization process

kirillmakhonin:
This change addresses Python classes serialization, that can raise an Exception on `__bool__` or `__eq__` operation. One of this specific-case classes is `pandas.DataFrame` (issue #261).

Way to reproduce:
```python
class RestrictedType:
    def __bool__(*args, **kwargs):
        raise Exception('Restricted function')
     __eq__ = __lt__ = __le__ = __ne__ = __gt__ = __ge__ = __hash__ = __bool__

 glob_obj = RestrictedType()

 def restricted_func():
    a = glob_obj

dill.loads(dill.dumps(restricted_func, recurse=True))
```

### DIFF
--- a/dill/_dill.py
+++ b/dill/_dill.py
@@ -1357,7 +1357,7 @@ def save_function(pickler, obj):
            #    if isinstance(value, stacktypes) and id(value) in stack:
            #        del globs[key]
             # ABORT: if self-references, use _recurse=False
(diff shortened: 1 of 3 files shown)


kirillmakhonin:
@mmckerns please have a look

PR #301, opened 2019-02-06
Fix shadowed namedtuple bug

vnetserg:
This fixes a case of incorrect unpickling of a namedtuple-derived class instance in a situation when the classes' module path has been shadowed by some other object (see the test provided in the commit for details).

/cc @Kontakter


mmckerns:
super corner case... thanks.

PR #302, opened 2019-02-06
Fix diff bug with oldstyle classes (Python 2)

vnetserg:
This fixes a bug in `__diff` module when it decides if an object is iterable or not. The problem is that it caches the result (`True` or `False`) for a type, but all the old-style class instances have the same type `<type 'instance'>`, so the first such an object that is passed to `get_seq` function determines the result for all the instances of all the old-style classes.

Also notice `#XXX` on `__diff.py:105` - maybe it has been added to alleviate the problem fixed in this commit?

/cc @Kontakter

### DIFF
--- a/dill/__diff.py
+++ b/dill/__diff.py
@@ -36,6 +36,12 @@ builtins_types = set((str, list, dict, set, frozenset, int))
 dont_memo = set(id(i) for i in (memo, sys.modules, sys.path_importer_cache,
              os.environ, id_to_obj))
 
+# InstanceType is a type of all the instances of old-style classes (Python 2).
+try:
+    InstanceType = types.InstanceType
+except AttributeError:
+    # Python 3
+    InstanceType = None
 
 def get_attrs(obj):
     """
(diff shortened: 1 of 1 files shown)


dill/__diff.py
```diff
@@ -57,7 +63,7 @@ def get_seq(obj, cache={str: False, frozenset: False, list: True, set: True,
     """
     Gets all the items in a sequence or return None
     """
-    o_type = type(obj)
+    o_type = type(obj) if type(obj) is not InstanceType else obj.__class__
```
matsjoyce:
```suggestion
    o_type = obj.__class__
```
Changing it to just `obj.__class__` should work?

vnetserg: I'm afraid this will fail on old-style class objects (`<type 'classobj'>`), which have no `__class__` attribute.

matsjoyce: `getattr(obj, "__class__", type(obj))` then? It would be nice to avoid the double call as diff is slow enough as it is...

vnetserg: Still it does double work: tries to get an attribute and also calls `type` regardless of the result. I guess the best we can do here performance-wise is
```
try:
    o_type = obj.__class__
except AttributeError:
    o_type = type(obj)
```

matsjoyce: I guess. Since we are supporting so many python versions I suppose there is only so much we can do.

vnetserg:
May I ask what's the status of the PR? I thought it can be merged as we have discussed the issue.

mmckerns:
The status is that it's been reviewed, but it hasn't been merged. I appreciate the PR @vnetserg, it'll get wrapped up shortly.

mmckerns:
Thanks @vnetserg

PR #320, opened 2019-06-11
Fixed the stuff to use setuptools better. 

KOLANICH:
Low-value stuff like inserting versions into README.md and adding a file with metainfo has been removed. If one needs a version info or content of README, he should use `pkg_resources`. Releases are the versions built from git tags. Development releases have git tags with "-dev" :) Also added some goodies, like .editorconfig and ignores.


setup.cfg
```diff
@@ -1,4 +1,47 @@
...
+    Programming Language :: Python :: 3.7
+    Topic :: Scientific/Engineering
+    Topic :: Software Development
+platforms = Linux, Windows, Mac
+
+[options]
+packages = dill, dill.tools
+zip_safe = False
+python_requires = >=2.5, !=3.0.*
+setup_requires = setuptools; wheel; setuptools_scm
```
mmckerns: While, in the next release or so, I will be trimming off support for some older versions of python, using `setuptools_scm` will force a drop of several old versions.  Old versions need to be supported longer in `dill` as compared to other packages, as people need a way to open old stored pickled object files.

setup.cfg
```diff
@@ -1,4 +1,47 @@
...
+
```
mmckerns: These two `cfg` are some nice modernization.

setup.py
```diff
@@ -6,314 +6,5 @@
 # License: 3-clause BSD.  The full license text is available at:
 #  - https://github.com/uqfoundation/dill/blob/master/LICENSE
 
-import os
-
-# set version numbers
```
mmckerns: Not sure I like the change in version numbering.  I need to ponder that a bit...

setup.py
```diff
@@ -6,314 +6,5 @@
...
-# check if easy_install is available
-try:
-#   import __force_distutils__ #XXX: uncomment to force use of distutills
-    from setuptools import setup
-    has_setuptools = True
-except ImportError:
-    from distutils.core import setup
-    has_setuptools = False
-
-# generate version number
```
mmckerns: Admittedly ugly (and hackish code from an old version of `numpy`) primarily to manage version number and docs.  I agree, it does need a change.

setup.py
```diff
@@ -6,314 +6,5 @@
...
-    if is_release:
-        from datetime import date
-        today = "".join(date.isoformat(date.today()).split('-'))
-        this_version += "-" + today
-
-# get the license info
-with open('LICENSE') as file:
-    license_text = file.read()
-
-# generate the readme text
```
mmckerns: This shouldn't go away.  I need these docs for ``dill.__init__.__doc__`.  Unfortunately they are similar but different than in `README.md`.   Not sure how to best resolve that.

setup.cfg
```diff
@@ -1,4 +1,47 @@
...
+    Programming Language :: Python :: 3.6
+    Programming Language :: Python :: 3.7
+    Topic :: Scientific/Engineering
+    Topic :: Software Development
+platforms = Linux, Windows, Mac
+
+[options]
+packages = dill, dill.tools
+zip_safe = False
+python_requires = >=2.5, !=3.0.*
```
mmckerns: `ctypes` requirement is missing.  It's needed for certain non-standard platforms, I believe like google cloud, if I remember correctly.

setup.cfg
```diff
@@ -1,4 +1,47 @@
...
+    Programming Language :: Python :: 3.4
+    Programming Language :: Python :: 3.5
+    Programming Language :: Python :: 3.6
+    Programming Language :: Python :: 3.7
+    Topic :: Scientific/Engineering
+    Topic :: Software Development
+platforms = Linux, Windows, Mac
+
+[options]
+packages = dill, dill.tools
```
mmckerns: I had `dill.tests` as a module.  While it might not be 100% desirable to include `tests` as a module, it at least was necessary for tests to run in some cases.  Needs investigation.

setup.py
mmckerns: The comments in the `More Info` section (also in `README.md`, I believe) are inconsistent with your changes.

setup.cfg
KOLANICH: isn't ctypes a part of python stdlih?

setup.py
KOLANICH: Of course it should go away. If we need `__init__.__doc__`, you can **probably** fetch the long_description using `pkg_resources` (I have never done that though).

setup.cfg
mmckerns: Yes... or more precisely...  believe it or not... most of the time it is.   The STL contents are not all the same on all platforms and distributions, and last I checked, it was one of a few libraries that isn't always shipped as part of the STL.

KOLANICH: >and ctypes doesn't come installed by default with pypy or if using MacPorts.

I guess that in this case it should be installed using system package manager, like `apt` and `brew`, not `pip`.

mmckerns: I can't control how people install `dill` -- the fact is sometimes the STL is slightly different due to how and where it was installed.

setup.py
mmckerns: You mean shouldn't, right?  `dill.__init__.__doc__` is the module doc, and is used for the `rtfd` entry page.  I typically see the `long_description` be similar if not the same as this doc.  Indeed that's what I've been doing.  Why remove this doc, and if you remove it, what do you put on the intro page for the sphinx docs?

KOLANICH: I mean the stuff bloating the module should go away - it is ugly and makes it uarder to read. And generating code in `setup.py` should go away. And custom `setup.py` for the most of projects should also go away. I mean we defilitely want to keep `setup.py` minimalist and clean and have everything in `setup.py`. This if sufficient count of projects adopted such minimalistic the same `setup.py` it should be possible to create a whitelist of hashes of such `setup.py` files and execute them only when they match the hash. This way it may be possible to get rid of remote code execution when installing python packages.

So IMHO if that large `__doc__` is really needed, it should be retrieved dynamically from metadata.

setup.cfg
KOLANICH: Thue problem is not with dill, but with its native dependencies which are meant to be a part of python interpreter.

setup.py
mmckerns: I agree `setup.py` could be cleaned up.  It has a lot of artifacts that were the way things were done idk maybe 3-5 years ago.  Further, having the doc in `setup.py` and having the doc in `__init__` are indeed two different things.  However, the `__doc__` is needed for the module documentation.  You've made it go away in this PR.  I'm not going to accept a PR that removes functionality unless it's really warranted, and I'm not at all convinced that removing module documentation is a good thing.  I'll consider your PR when if you decide you want to handle the review comments.

setup.cfg
mmckerns: I agree.  But, we have a practical problem here, about what to do about it.

setup.py
KOLANICH: I am not going to fix anything in my PR. The main motivation was to clear the disgusting mess in `setup.py` because I hate the mess. Feel free to create an own commit based on this one doing all the fixes you like.

mmckerns:
there's been significant updates to dill's build procedure. It still might be nice to cherrypick some stuff from this PR.

Issue #323, opened 2019-06-30
`dill` fails to access global dict on windows


mmckerns:
Using `multiprocess` with `dill-0.3.0` on windows, it seems that `dill` fails to access the global dict... however, older versions of `dill` (e.g. `0.2.x`) work as expected.  These errors are not seen on non-windows systems.  See original reporting here:
See: https://github.com/uqfoundation/multiprocess/issues/61

mmckerns:
Also see: https://github.com/uqfoundation/multiprocess/issues/65 for some possibly related context

molinav:
I also include a link to [this comment in the pathos issue tracker](../../pathos/issues/118#issuecomment-617870380), I think that the `NameError` that I get in my minimal working example is triggered by this issue.

In my case, the error also appears with dill 0.2.9 on Windows. If I downgrade dill to 0.2.8.2, then the error disappears.

molinav:
Just for the record, I cloned the repo and tested my minimal working example for all the commits in the `master` branch between tag `dill-0.2.8.2` and tag `dill-0.2.9`. Starting with commit 3b043d67 the example fails.

molinav:
Let's copy here the minimal working example that I was mentioning:
```python
# test_example_multiprocess.py


def func1():
    print("Hello world!")


def func2():
    func1()


if __name__ == "__main__":

    from multiprocess import Process
    proc = Process(target=func2)
    proc.start()
    proc.join()
```
```sh
(py37) D:\vic\Desktop>python test_example_multiprocess.py
Process Process-1:
Traceback (most recent call last):
  File "C:\GNU\Anaconda3\envs\py37\lib\site-packages\multiprocess\process.py",
  line 297, in _bootstrap
    self.run()                                                                      
  File "C:\GNU\Anaconda3\envs\py37\lib\site-packages\multiprocess\process.py",
  line 99, in run
    self._target(*self._args, **self._kwargs)
  File "test_example_multiprocess.py", line 9, in func2
    func1()
NameError: name 'func1' is not defined

(py37) D:\vic\Desktop>
```

The function that takes care of serialising the module variables is [`save_module_dict`]. If I print the `obj` dictionary for the file above when entering [`save_module_dict`], it looks like below:
```python
{
    '__name__': '__main__',
    '__doc__': None,
    '__package__': None,
    '__loader__': <_frozen_importlib_external.SourceFileLoader object at 0x000002BC66F1C108>,
    '__spec__': None,
[...]

PR #363, opened 2020-04-26
Patch module variables pickling on Windows (#323)

molinav:
This pull request includes the `child` argument explicitly in all the calls to the function `is_dill`. This argument is set to `True` in all of them except for those calls inside `save_module_dict`, in order to patch the error that is triggered on Windows as described in #323.


mmckerns:
thanks

Issue #386, opened 2020-10-23
Functions defined with exec() are missing their environment w/ recurse=False; relevant doc needed


pkienzle:
I'm unable to call a function defined by `exec()` after serializing and deserializing if the function has free variables.

To reproduce:
```python
filename = "check.py"
source = """
value = 1
def f():
    return value
"""
context = {}
code = compile(source, filename, 'exec')
exec(code, context)
print("direct", context['f']())

import dill
print("copied", dill.loads(dill.dumps(context['f']))())
```
shows:
```
direct 1
Traceback (most recent call last):
  File "dillcheck.py", line 15, in <module>
    print("local", dill.loads(dill.dumps(context['f']))())
  File "check.py", line 4, in f
NameError: name 'value' is not defined
```

This may be related to #145 but the `__builtins__` workaround suggested there did nothing.

pkienzle:
Adding `recurse=True` resolves the error.

I'm leaving the issue open since it is a documentation problem.

`dumps` has a couple of arguments that are not part of the usual pickle interface, and I wasn't able to find a description in the documentation:
* `byref`
* `fmode`
* `recurse`
* `**kwds`

Similarly, `loads` has:
* `ignore`
* `**kwds`

mmckerns:
@pkienzle!  Thanks for the ping.  Yeah, I'd close this issue as a duplicate, and add the comment to #145 to try one of the serialization variants (i.e. `recurse`)... but I think if you don't mind, what I'm going to do is change the title of this ticket to reflect that there are documentation issues... and then leave this open until they are resolved.  The undocumented settings are "per-function call" serialization variants -- so `recurse` is a variant that works a lot like `cloudpickle`.

pkienzle:
This makes it a dup of #122 so one or another can be closed.

pkienzle:
I harvested some of your comments and what I could glean from the code and threw together some basic docs.

I'm not really sure when you need `recurse`, `byref` or `ignore`, or why you would choose `fmode=CONTENT_FMODE` over `fmode=FILE_FMODE`, but hopefully what I wrote helps more than it confuses.

I dropped in a couple of FIXME's, though nothing of consequence (I don't think your tests for numpy vector equality does what you intended for example).

pkienzle:
[...]

Issue #122, opened 2015-08-07
Add documentation for options (byref, fmode, recurse, etc)


TD22057:
It would be really helpful if there was some documentation for all of the options to dump().  Please add a couple of lines to explain the options to the function doc strings.  Ideally, the users guide would then expand on that to include more examples of when you might want to use each flag.

matsjoyce:
It would be cool to use sphinx for this, because then we can make web accessible / readthedocs / pdf documentation as well.

mmckerns:
Yes, I agree, the settings should be documented better -- at least in the module level documentation, and at most in every method in which they are used.

I did at one time (2010 and earlier) keep `epydoc` pages for `dill` and my other codes, so a lot of the work to make the module docs `reST` compliant is already done.  `sphinx` was not yet in wide usage at the time, so I went with `epydoc`. See: http://docs.danse.us/pathos/current/dill-module.html.

I'm usually pretty good about having good in-code documentation that is exposed to the user through `help`… so the lack of `dill.settings` doc is a bit of an oversight.  I've never been very good about writing additional tutorial docs, so anything in that vein I could use a bit of help on.  There's probably a lot of good examples that could be pulled from stackoverflow, however.

I'm open to `sphinx` docs.  It's probably the right thing to do… but definitely the `dill.settings` stuff should be documented better as a first step.

robert-simione:
Found this ticket, just wanted to add that documentation on common flags would have saved over a day of digging if the behavior of dill.settings['recurse'] was documented somewhere. I only found out that I needed this setting for the behavior I expected from dill after stumbling into this question https://github.com/uqfoundation/dill/issues/188

spenceryee:
Any update on this? I've been digging around and I'm still not entirely sure what `recurse` does.

mmckerns:
No, the module-level docs still need to be written.  In short, `byref`, `recurse` are serialization variants.  If `byref = True`, then `dill` behaves a lot more like `pickle` as certain objects (like modules) are pickled by reference as opposed to attempting to pickle the object itself.  If `recurse = True`, then `dill` behaves a lot more like `cloudpickle` in that objects referred to in the global dictionary are recursively traced and pickled, instead of the default behavior of attempting to store the entire global dictionary.

mmckerns:
Closing due to #388

PR #388, opened 2020-10-24
DOC: add basic usage info and description of serialization options

pkienzle:
Refs #122, #386

### DIFF
--- a/dill/_dill.py
+++ b/dill/_dill.py
@@ -89,7 +89,7 @@ if sys.hexversion < 0x03030000:
     FileNotFoundError = IOError
 if PY3 and sys.hexversion < 0x03040000:
     GENERATOR_FAIL = True
-else: GENERATOR_FAIL = False    
+else: GENERATOR_FAIL = False
 if PY3:
     import importlib.machinery
     EXTENSION_SUFFIXES = tuple(importlib.machinery.EXTENSION_SUFFIXES)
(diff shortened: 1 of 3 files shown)


Azmisov:
👍  Let's merge this. Those docs need some love

dill/_dill.py
```diff
@@ -268,17 +276,52 @@ def dump(obj, file, protocol=None, byref=None, fmode=None, recurse=None, **kwds)
...
 def dumps(obj, protocol=None, byref=None, fmode=None, recurse=None, **kwds):#, strictio=None):
-    """pickle an object to a string"""
+    """
+    Pickle an object to a string.
+
+    *protocol* is the pickler protocol, as defined for Python *pickle*.
+
+    If *byref=True*, then dill behaves a lot more like pickle as certain
+    objects (like modules) are pickled by reference as opposed to attempting
+    to pickle the object itself.
```
rainwoodman, line 286: Thanks for the example on module under byref=True, but I think this section may need some additional elaboration: 

How would byref=True affect the pickling of classes, functions?
What about classes and functions created in a nested scope?
What about dynamically created modules without a source file?

What is the definition of a 'ref? Does 'ref' here stand for a Python fully qualified name to an object such as ("package.module.Class.NestedClass.member")?

Would we expect an error when byref=True and the object cannot be constructed from the reference?

mmckerns (approved):
Thanks Paul.  FYI, there's a small doc bug, but I'll fix it with a patch commit.

Issue #394, opened 2020-12-03
No module named 'dill.dill' when using pickle.load


tayciryahmed:
Hi, 

I am trying to load in python3.7 a file -that was serialised using `dill.dump` in python2.7.
I am following this tutorial: https://rebeccabilbro.github.io/convert-py2-pickles-to-py3/

```
import dill, pickle 
dill._dill._reverse_typemap["ObjectType"] = object
with open(serialized-file, "rb") as f:
    loaded = pickle.load(f, encoding="latin1")

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'dill.dill'

```

```
dill.__version__
'0.3.3'
```

Any ideas on how to solve this ?

mmckerns:
This is unsupported.  It can work... but it's unsupported.  First off, it appears that the object that was serialized was with an old `dill`, which moved from `dill.dill` to `dill._dill`, as you surmised.  You could try to rename the `_dill.py` file to `dill.py` and go with a more recent version of `dill`... but I have no idea if that would work.  Essentially, you are going to be dealing with every incompatible difference between python2 and python3 as the objects are unpickled, and you'll need to manually convert the python2 object to a python3 object by registering a custom deserialization method in the `typemap`.  Not impossible, but difficult and tedious.   Good luck.

tayciryahmed:
Thanks for the quick reply. 
I decomposed my picked pipeline to track the object that is causing this error `TypeError: an integer is required (got type str)`. I found out that pickling and un-pickling the custom classes causes the same error. So, I have tried the following steps:

**In python2.7 and dill==0.3.3**:

```
import dill

class test(object):
    def __init__(self): pass
    def fit(self, x): return x

t = test()

f = open('test', 'wb')
dill.dump(t, f)
f.close()
```

**Then in python3.7 and dill==0.3.3:**

```
import dill 

dill._dill._reverse_typemap["ClassType"] = type
dill._dill._reverse_typemap["TypeType"] = type
dill._dill._reverse_typemap["ObjectType"] = object

t = dill.load(open('test', 'rb'))
t.fit(1)

XXX lineno: 3, opcode: 0
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "<stdin>", line 3, in fir
SystemError: unknown opcode
```

I get the above error `SystemError: unknown opcode`. 

Any ideas on how to debug this?

Issue #357, opened 2020-02-05
Python 3.8 compatibility issues TypeError


Huyston:
A similar problem was reported in #318, but it was closed.

However, I still have this issue with python 3.8.1 and dill 0.3.1.1

```
Traceback (most recent call last):
  File "main.py", line 988, in openFile
    self.archive[docID] = pickle.load(archive)
  File "/usr/lib/python3.8/site-packages/dill/_dill.py", line 270, in load
    return Unpickler(file, ignore=ignore, **kwds).load()
  File "/usr/lib/python3.8/site-packages/dill/_dill.py", line 472, in load
    obj = StockUnpickler.load(self)
TypeError: an integer is required (got type bytes)
```

metya:
Confirm!

GGooeytoe:
To replicate (using conda python on Mac OSX)
in python 3.7.6, with dill version 0.3.1.1, do:
```
import dill
foo=lambda x:x
with open("foo.pkl","wb") as fh:
    dill.dump(foo,fh)
```

in python 3.8.1, with dill version 0.3.1.1, do:
```
import dill
with open("foo.pkl","rb") as fh:
    foo=dill.load(fh)
```

Result:
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-2-a1a2042a49f3> in <module>
      1 with open("foo.pkl","rb") as fh:
----> 2     foo=dill.load(fh)
      3

~/miniconda3/envs/uq2/lib/python3.8/site-packages/dill/_dill.py in load(file, ignore, **kwds)
    268 def load(file, ignore=None, **kwds):
    269     """unpickle an object from a file"""
--> 270     return Unpickler(file, ignore=ignore, **kwds).load()
    271
    272 def loads(str, ignore=None, **kwds):

~/miniconda3/envs/uq2/lib/python3.8/site-packages/dill/_dill.py in load(self)
    470
    471     def load(self): #NOTE: if settings change, need to update attributes
--> 472         obj = StockUnpickler.load(self)
    473         if type(obj).__module__ == getattr(_main_module, '__name__', '__main__'):
    474             if not self._ignore:

TypeError: an integer is required (got type bytes)
```

This also happens with python3.6.6->python3.8.1

nirvana-msu:
[...]

Issue #318, opened 2019-06-05
Python 3.8 issue for code objects due to position-only arguments


irmen:
Dill has at least the following compatibility problem with Python 3.8  (which just released the first beta version):

```
Python 3.8.0b1+ (heads/3.8:d74438b633, Jun  5 2019, 22:17:23) 
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import dill
>>> def func(x): pass
... 
>>> p=dill.dumps(func)   
>>> dill.loads(p)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/irmen/venv/py38/lib/python3.8/site-packages/dill/_dill.py", line 317, in loads
    return load(file, ignore)
  File "/home/irmen/venv/py38/lib/python3.8/site-packages/dill/_dill.py", line 305, in load
    obj = pik.load()
TypeError: an integer is required (got type bytes)
>>> 

```

mmckerns:
Yuck.  Thanks for the ping.  I was waiting for the beta to drop, and apparently it has.

irmen:
Indeed it has! I was using it to validate my own projects against it.  One test failed due to the dill dependency.

mmckerns:
I see, the issue above is due to changes to accommodate position-only functions.

mmckerns:
The above PR fixes this issue.  Thanks for the poke. Please open another PR if you find something else.

mmckerns:
@irmen: new release was cut earlier this week.  I'll continue to work on any issues that arise with changes from 3.8.  All tests pass, but I'm sure there are issues lurking.

irmen:
@mmckerns can confirm green status on Pyro4's python 3.8 build :) thanks!

metya:
I have this error on my `python 3.8.1` and `dill 0.3.1.1`

Huyston:
I still have this issue with python 3.8.1 and dill 0.3.1.1

```
Traceback (most recent call last):
  File "main.py", line 988, in openFile
    self.archive[docID] = pickle.load(archive)
  File "/usr/lib/python3.8/site-packages/dill/_dill.py", line 270, in load
    return Unpickler(file, ignore=ignore, **kwds).load()
  File "/usr/lib/python3.8/site-packages/dill/_dill.py", line 472, in load
    obj = StockUnpickler.load(self)
TypeError: an integer is required (got type bytes)
```

irmen:
[...]

PR #406, opened 2021-03-12
Allow 3.7 Pickles to be Loaded in 3.8

njwhite:
Use `_create_code` Logic when loading pickled objects rather than just the builtin `CodeType`. If the pickle file was created using, e.g., Python 3.7 then the serialized object will 15 arguments (missing `co_posonlyargcount`) but the version in the current (Python 3.8) interpreter expects 16. This PR just fills in a zero.

Related issues: #357 #318 #394 cloudpipe/cloudpickle#396

### DIFF
--- a/dill/_dill.py
+++ b/dill/_dill.py
@@ -534,6 +534,23 @@ def use_diff(on=True):
             import diff as d
         diff = d
 
+def _create_code(*args):
+    if PY3 and hasattr(args[-3], 'encode'): #FIXME: from PY2 fails (optcode)
+        args = list(args)
+        args[-3] = args[-3].encode() # co_lnotab
+        args[-10] = args[-10].encode() # co_code
+    if hasattr(CodeType, 'co_posonlyargcount'):
+        if len(args) == 16: return CodeType(*args)
+        elif len(args) == 15: return CodeType(args[0], 0, *args[1:])
+        return CodeType(args[0], 0, 0, *args[1:])
+    elif hasattr(CodeType, 'co_kwonlyargcount'):
+        if len(args) == 16: return CodeType(args[0], *args[2:])
+        elif len(args) == 15: return CodeType(*args)
+        return CodeType(args[0], 0, *args[1:])
+    if len(args) == 16: return CodeType(args[0], *args[3:])
+    elif len(args) == 15: return CodeType(args[0], *args[2:])
+    return CodeType(*args)
+
 def _create_typemap():
     import types
     if PY3:
@@ -565,6 +582,7 @@ _reverse_typemap.update({
     'PyBufferedReaderType': PyBufferedReaderType,
     'PyBufferedWriterType': PyBufferedWriterType,
     'PyTextWrapperType': PyTextWrapperType,
+    'CodeType': _create_code,
 })
 if ExitType:
     _reverse_typemap['ExitType'] = ExitType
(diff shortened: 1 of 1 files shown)


njwhite:
Hi @mmckerns - can I get a review for this? I’m just moving around a few lines of existing code so I don’t think the code coverage CI failure is meaningful. Thanks -

mmckerns:
Impact of the change needs to be assessed on the various use cases.

mmckerns:
Is your description still correct?  It seems that all this PR does is to move  `_create_code` and then add it to the `typemap`.  Can you clarify what is this PR meant to do, exactly?

njwhite:
The key bit is line 585 - when the deserialisation code sees a CodeType it now uses the backcompat logic in _create_code to interpret it. Previously dill would just try to read the CodeType using the default logic, which fails on 3.8 when you try to read a 3.7 pickle as the signature has changed.

mmckerns:
Here's an example of the current behavior:
```
Python 3.7.10 (default, Mar 18 2021, 06:11:04) 
[Clang 9.0.0 (clang-900.0.39.2)] on darwin
>>> import dill
>>> def foo(w, x=1, *y, **z):
...   return w+x+sum(y)+sum(z.values())
... 
>>> f = open('foo.pkl', 'wb')
>>> dill.dump(foo, f)
>>> dill.dump(foo.__code__, f)
>>> f.close()
```
and then in 3.8
```
Python 3.8.10 (default, May  7 2021, 23:18:56) 
[Clang 9.0.0 (clang-900.0.39.2)] on darwin
>>> import dill
>>> f = open('foo.pkl', 'rb')
>>> foo = dill.load(f)
>>> c = dill.load(f)
>>> foo(1)
2
>>> f.close()
```
Can you provide a case where `dill` currently fails, but your PR enables it to succeed?

njwhite:
@mmckerns I've pushed a test case - it fails if you comment out the [the handler](https://github.com/uqfoundation/dill/pull/406/files#diff-bdd40cc870892c410adce8c1b155daf5e6c5b5e7192e751f171e6dae6a578dc1R629).

tests/test_functions.py
```diff
@@ -30,6 +31,14 @@ def function_d(d, d1, d2=1):
     return d + d1 + d2
 
 
+def lambda_a():
+    pkl = os.path.join(
+        os.path.dirname(__file__),
+        "lambda.pkl")
```
mmckerns, line 37: It would be much better to write the file contents elsewhere, instead of relying on a stored pickle file.  Was the file written in python 3.7?  `dill` tests are currently run with 2.7, 3.6, 3.7, 3.8, 3.9, 3.10, pypy27, pypy36, and pypy37.  Is the test only supposed to run with 3.8?

You don't need to add a test case into the code at the moment, if it's difficult.  Rather, just present the details in the main conversation of the Github issue, so I and others can reproduce what you are seeing.

njwhite: The file was written with python 3.7 /  dill 0.3.0. I think that’s why you can’t put the file contents there (and need the binary) - the bug seems to be that dill isn’t backwards compatible. 

The pickled object is just lambda x: x.

mmckerns: Here's an example using lambda in `dill` master with 3.7:
```
Python 3.7.10 (default, Mar 18 2021, 06:11:04) 
[Clang 9.0.0 (clang-900.0.39.2)] on darwin
>>> import dill
>>> dill.dump(lambda x:x, open('test.pkl', 'wb'))
```
and loading with 3.8...
```
Python 3.8.10 (default, May  7 2021, 23:18:56) 
[Clang 9.0.0 (clang-900.0.39.2)] on darwin
>>> import dill
>>> f = dill.load(open('test.pkl', 'rb'))
>>> f(4)
4
>>> 
```
What is your PR doing that is not possible currently?

njwhite: You need to dump with dill 0.3.0 and python 3.7, not master/3.7 to reproduce.

mmckerns: So... you are saying that the issue is that old pickles from python 3.7 created with `dill` 0.3.0 don't unpickle in python 3.8 with `dill` master.  I'm assuming this is also the case for other old versions of `dill` (before `_create_function`  was recently modified).

Am I correct in thinking that you could, as a workaround, load the pickle in 3.7 with `dill` master, and then dump it again... then the resulting file would be able to be opened with 3.8?

njwhite: My 3.8 workaround is just to set:

 dill._dill._reverse_typemap[‘CodeType’] = dill._dill._create_code

much easier than re-serialising all the pickle files I have lying around :)

mmckerns: Riiight, of course... hence the PR.  I never really considered it, however, I'm wonder if adding other of the `_create_` functions to the `reverse_typemap` is worth investigating.  I'm not certain of what functionality it might impact.

PR #421, opened 2021-07-17
pickle function: fattributes

NickFillot:
Pickle pyfunction attributes like __annotations__ / Pandas DataFrame


mmckerns:
see #422

PR #448, opened 2022-01-27
Pickle inner `collections.namedtuple`s and function attributes

### LINKED ISSUE #288: Dill gives incorrect, invalid name to nested namedtuples in 3.7
This (reduced from real-world code) example fails:

    import collections, dill

    class A:
        class B(collections.namedtuple("B", ["one", "two"])):
            pass
    
    a = A()
    aa = dill.loads(dill.dumps(a))

Dill thinks `B` is named `A.B`, which is not a valid name for a NamedTuple since it has to be a valid identifier, and Python gives `ValueError: Type names and field names must be valid identifiers: 'A.B'`

Adding the following makes it work again for me at least, but I don't know whether or not the name being wrong should be considered a deeper issue.

In _dill._create_namedtuple:

    import collections
    name = name.split('.')[-1] #add this line - removes the parent's name, 'A.B' -> 'B'
    t = collections.namedtuple(name, fieldnames)

### LINKED ISSUE #420: [Enhancement] pickle function __annotations__
Hello,

Is it possible to implement function.__annotations__ pickle on loads / dumps ?

Br
Nick

anivegesana:
1) Fixing pickling of `collections.namedtuple` subclasses
2) Fixes pickling `collections.namedtuple` when `__qualname__` is not `__name__`
3) Saves nonessential attributes of function objects by name instead arguments to the `_create_function` (supersedes #422) 
4) Pickles defaults of `collections.namedtuple` classes

I think it is worth debating if this is the best way to handle new attributes to the function class. Some functions have `__annotations__` but not `__kwdefaults__`. In the future, if these new attributes don't scale well, the current solution of adding them to `_create_function` could likely lead to a mess. On the other hand, these new strings would be added to all pickles whose functions that use said attributes. When many functions are pickled in one pickle, this overhead is constant. However, when pickling many functions in separate pickles, the overhead becomes linear in the number of functions. This cost should be considered.

This PR also adds a breaking change. Previously, `_create_namedtuple` did not save and use the default values of the tuple. In order to add the argument to this function to support it, pickling a `collections.namedtuple` with default values using this PR and trying to unpickle it in an older version of dill will fail.

Fixes #288, Fixes #420

### DIFF
--- a/dill/_dill.py
+++ b/dill/_dill.py
@@ -707,9 +707,14 @@ def _create_function(fcode, fglobals, fname=None, fdefaults=None,
                      fclosure=None, fdict=None, fkwdefaults=None):
     # same as FunctionType, but enable passing __dict__ to new function,
     # __dict__ is the storehouse for attributes added after function creation
-    if fdict is None: fdict = dict()
     func = FunctionType(fcode, fglobals or dict(), fname, fdefaults, fclosure)
-    func.__dict__.update(fdict) #XXX: better copy? option to copy?
+    if fdict is not None:
+        func.__dict__.update(fdict) #XXX: better copy? option to copy?
+    elif IS_PYPY2:
+        # __reduce__ crashes in PyPy2
+        # __setstate__ is not used in any PyPy2 code and is removed in PyPy3
+        # so is likely an artifact
+        func.__setstate__ = None
     if fkwdefaults is not None:
         func.__kwdefaults__ = fkwdefaults
     # 'recurse' only stores referenced modules/objects in fglobals,
@@ -971,14 +976,23 @@ def _create_dtypemeta(scalar_type):
         return NumpyDType
     return type(NumpyDType(scalar_type))
 
-def _create_namedtuple(name, fieldnames, modulename):
-    class_ = _import_module(modulename + '.' + name, safe=True)
-    if class_ is not None:
-        return class_
-    import collections
-    t = collections.namedtuple(name, fieldnames)
-    t.__module__ = modulename
-    return t
+if OLD37:
+    def _create_namedtuple(name, fieldnames, modulename, defaults=None):
+        class_ = _import_module(modulename + '.' + name, safe=True)
+        if class_ is not None:
+            return class_
+        import collections
+        t = collections.namedtuple(name, fieldnames)
+        t.__module__ = modulename
+        return t
+else:
+    def _create_namedtuple(name, fieldnames, modulename, defaults=None):
+        class_ = _import_module(modulename + '.' + name, safe=True)
+        if class_ is not None:
+            return class_
+        import collections
+        t = collections.namedtuple(name, fieldnames, defaults=defaults, module=modulename)
+        return t
 
 def _getattr(objclass, name, repr_str):
     # hack to grab the reference directly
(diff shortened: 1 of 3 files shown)


anivegesana:
I believe that the only issue for this PR is coverage. Some cases apply to Python 3.6 where defaults are not available for `collections.namedtuple` and some for Python 3.7+ when this feature was added.

mmckerns:
Whenever you are ready for a review (on this or other of your PRs), just request me as a reviewer.

anivegesana:
I think that requires permissions, just like merging an already approved PR. The open PRs are ready to review. I am still refactoring the draft, but it is functional.

anivegesana:
Looking back on it, I think that #461 might be a little incomplete. I think that I need to ensure that every slot defined under "User-defined functions" on https://docs.python.org/3/reference/datamodel.html gets copied over when it isn't already the default value.

anivegesana:
Looking back on it, I think that #461 might be a little incomplete. I think that I need to ensure that every slot defined under "User-defined functions" on https://docs.python.org/3/reference/datamodel.html gets copied over when it isn't already the default value.

anivegesana:
This PR seems to be backwards incompatible for PyPy 2.7. A function pickled in this PR will be unable to be unpickled in dill 0.3.4 on PyPy 2.7, but will be able to be unpickled on all other versions. Should there be a setting to turn off this feature to allow for the creation of backwards compatible pickles?

mmckerns:
> This PR seems to be backwards incompatible for PyPy 2.7. A function pickled in this PR will be unable to be unpickled in dill 0.3.4 on PyPy 2.7, but will be able to be unpickled on all other versions. Should there be a setting to turn off this feature to allow for the creation of backwards compatible pickles?

Ouch. No. The good thing is that it's only for PyPy 2.7.  I'll have to have a look and see if there's any way to avoid the incompatibility.

anivegesana:
There is an easy way. Instead of using `__setstate__`, I unroll it into a chain of `setattr` calls. It is just more inefficient. That is why a setting for this might be useful.

mmckerns:
What's done with python 2.7 doesn't matter too much.  I will be dropping support for 2.7 immediately after the next release.  With regard to what to do, I'd rather not lose functionality from one release to the next -- and would much prefer living with a small hit on the time it takes to pickle something in 2.7.  So, yes, the overhead of a `setattr` call is preferred.

anivegesana:
Got it.

mmckerns:
There's an issue if the `dill` pickler is not used.  I typically check `is_dill` to guard against missing attributes.
```
$ python test_readwrite.py 
  File "./test_readwrite.py", line 144, in <module>
    test_archive()
  File "./test_readwrite.py", line 122, in test_archive
    check_basic(archive)
  File "./test_readwrite.py", line 80, in check_basic
    d['d'] = squared
  File "/Users/mmckerns/lib/python3.7/site-packages/klepto-0.2.2.dev0-py3.7.egg/klepto/_archives.py", line 397, in __setitem__
    self._store(key, value, input=False) # input=True also stores input
  File "/Users/mmckerns/lib/python3.7/site-packages/klepto-0.2.2.dev0-py3.7.egg/klepto/_archives.py", line 621, in _store
    protocol=protocol)
  File "/Users/mmckerns/lib/python3.7/site-packages/klepto-0.2.2.dev0-py3.7.egg/klepto/_pickle.py", line 389, in dump
    pickler.dump(value)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/pickle.py", line 437, in dump
    self.save(obj)
  File "/Users/mmckerns/lib/python3.7/site-packages/klepto-0.2.2.dev0-py3.7.egg/klepto/_pickle.py", line 264, in save
    return Pickler.save(self, obj)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/pickle.py", line 504, in save
    f(self, obj) # Call unbound method with explicit self
  File "/Users/mmckerns/lib/python3.7/site-packages/dill-0.3.5.dev0-py3.7.egg/dill/_dill.py", line 1921, in save_function
    topmost_postproc = next(iter(pickler._postproc.values()), None)
AttributeError: 'NumpyPickler' object has no attribute '_postproc'

```

anivegesana:
Good catch. I guess I forgot to copy that code from the merge commit.

mmckerns (approved):
LGTM

PR #449, opened 2022-01-27
Support `dict` views and `functools.lru_cache`

### LINKED ISSUE #353: Cannot pickle dict_items, dict_keys & dict_values on Python-3
This simple code fails on Python3 9but on Python 2):
```python
dill.dumps({}.items())
```

It is a pretty common object in the python universe.
.

anivegesana:
1) Implement save functions for dictionary view objects and LRU caches
2) Add trick to extract the dictionary directly from a `types.MappingProxyType` in CPython 3.9+
3) Use trick to preserve reference structure for dictionary view objects in Python 3.10+

Fixes #353

### DIFF
--- a/dill/_dill.py
+++ b/dill/_dill.py
@@ -40,6 +40,7 @@ PY3 = (sys.hexversion >= 0x3000000)
 OLDER = (PY3 and sys.hexversion < 0x3040000) or (sys.hexversion < 0x2070ab1)
 OLD33 = (sys.hexversion < 0x3030000)
 OLD37 = (sys.hexversion < 0x3070000)
+OLD310 = (sys.hexversion < 0x30a0000)
 PY34 = (0x3040000 <= sys.hexversion < 0x3050000)
 if PY3: #XXX: get types from .objtypes ?
     import builtins as __builtin__
@@ -261,6 +262,8 @@ except NameError:
     try: ExitType = type(exit) # apparently 'exit' can be removed
     except NameError: ExitType = None
     singletontypes = []
+
+from collections import OrderedDict
 
 import inspect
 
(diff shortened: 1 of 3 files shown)


mmckerns (approved):
LGTM

anivegesana:
A kind of unrelated side note (can definitely be ignored until after the release and is probably unimportant since I don't think anyone complained about it), but I think a handful of classes are missing from `_reverse_typemap`. Should I add them to the dictionary and what naming scheme should I give them? These are the classes in CPython. In PyPy, most of the iterators are handled by just two classes: `sequenceiterator` and `reversesequenceiterator`, so there will be duplicates in the map.
```
x=collections.OrderedDict()


"bytes_iterator": type(iter(b'')),
"bytearray_iterator": type(iter(bytearray(b''))),
"callable_iterator": type(iter(iter, None)),
"memory_iterator": type(iter(memoryview(b''))),
"range_iterator": type(iter(range(3))),
"set_iterator": type(iter(set())),
"tuple_iterator": type(iter(())),

"dict_keys": type({}.keys()),
"dict_values": type({}.values()),
"dict_items": type({}.items()),

"dict_keyiterator": type(iter({}.keys())),
"dict_reversekeyiterator": type(reversed({}.keys())),
"dict_valueiterator": type(iter({}.values())),
"dict_reversevalueiterator": type(reversed({}.values())),
"dict_itemiterator": type(iter({}.items())),
"dict_reverseitemiterator": type(reversed({}.items())),

"list_iterator": type(iter([])),
"list_reverseiterator": type(reversed([])),

"odict_keys": type(x.keys()),
"odict_values": type(x.values()),
"odict_items": type(x.items()),

"odict_iterator": iter(x.keys()),

if CPython: "symtable_stentry_type": type(symtable.symtable("", "string", "exec")._table),

if Windows: "PyHKEY": winreg.HKEYType,
```

mmckerns:
> A kind of unrelated side note...

@anivegesana: Feel free to open a new issue for the above comment. I'm sure that I've missed adding types to the reverse typemap over the years, so it will be good to add any that I've left out.  I generally try to use a name that is similar to whatever name the type identifies itself as.  You can see a bunch of examples in `dill/_objects.py`.

PR #463, opened 2022-04-22
Fixes some bugs when using `dump_session()` with `byref=True`

### LINKED ISSUE #462: `load_session()` fails when `dump_session()` is used with `byref=True`
`load_session()` is failing even in the simplest cases when `dump_session()` was used with the `byref` parameter set to `True`:

```python
>>> import dill
>>> x = 1
>>> dill.dump_session(byref=True)
# restart the interpreter, or not
>>> x = 0
>>> dill.load_session()
>>> x
0
# should have updated 'x' value
>>> del x
>>> dill.load_session()
  File "/home/leogama/.local/lib/python3.8/site-packages/dill/_dill.py", line 423, in load_session
    _restore_modules(main)
  File "/home/leogama/.local/lib/python3.8/site-packages/dill/_dill.py", line 383, in _restore_modules
    exec("from %s import %s" % (module, name), main_module.__dict__)
  File "<string>", line 1, in <module>
ImportError: cannot import name 'x' from '__main__' (unknown location)
```

Am I missing something? I tried to take a look at what happens differently when the `byref` option is used, but I don't understand the manipulations made to the `__main__` module.

---

My setup:
- Ubuntu 20.04.4 LTS
- Linux 5.4.0-107
- Python 3.8.10
- dill 0.3.4

leogama:
Here are the proposed fixes for bugs reported in https://github.com/uqfoundation/dill/issues/462.

You can review the changes already, but **please don't merge it yet** as I'm still tracking a strange third bug in `dump_session()`. And I spotted yet another possible bug, but need to test it first. I'll probably append more commits in the next days.

I'm working on this feature for it to be used as the cache mechanism for the `reticulate` R package —`knitr`'s Python engine. It's mostly functional by now. (Related discussion: https://github.com/yihui/knitr/issues/1505)

By the way, could you clarify a thing for me? Why is it that, even when the `byref` parameter is set to `True`, after the imported objects are identified and stored in the `__dill_imported` object, the `Pickler` is called with the `_byref` attribute set to `False`. What's the rationale? Is it because the session is saved as a whole module and then `pickle` would try to store everything inside it by name reference?

Fixes #462


mmckerns:
Yes, if we didn't set `_byref=False` while trying to store the main module, it will attempt to use name reference for modules, etc.

The PR looks good thus far. FYI, I expect that we will create a new release this coming week.

mmckerns:
Yes, if we didn't set `_byref=False` while trying to store the main module, it will attempt to use name reference for modules, etc.

The PR looks good thus far. FYI, I expect that we will create a new release this coming week.

leogama:
> The PR looks good thus far. FYI, I expect that we will create a new release this coming week.

This are great news! It would be awesome to have `dump_session(byref=True)` working fine in this next release. I will definitely need it to be able to propose changes in the (maybe) two R packages related to the usage of `dill` as a cache engine for Python in R Markdown documents. Do you mind to include this PR in the 0.3.5 milestones? I'm happy if any number of these bug fixes are included, and will continue to work on this feature in a new PR after the release.

mmckerns:
if we can finish the PR, it will be included in the release

leogama:
Hello there! My PR is ready for review. I rushed to solve all the problems I've found using `dump_session()`, so it could be included in the next release. The overall result seems to be satisfactory and working in both python2 and python3 (3.8 on my machine). I still don't understand fully the function saving logic, but contrary to my first belief it didn't need to be modified.

Do we need some unit tests? How can I provide them? Should it just reproduce the kinds of usage that were failing before these modification? I'm new to testing.

mmckerns:
Cool. I'll review it tonight.  Yes, `dump_session` sorely needs unit tests.  Most of `dill` has good unit test coverage, however `dump_session` may have none or at least very very few.  Testing is done in `dill.tests`.  You are free to add a module in `dill.tests`, if one doesn't exist that tests `dump_session`.  Then, you should include yourself as an author of the file, as an additional line to the standard file header.

leogama:
Thank you! I'll take a look at the tests.

mmckerns:
tests have to pass for the currently supported versions of python which are: (python 2.7, 3.7-3.10; PyPy 2.7, 3.7-3.9) and should also pass for python 3.11.  We will drop support for python 2.7 / PyPy 2.7 after this upcoming release.

mmckerns:
I did some random testing -- building a class and an instance of a class, then dumping a session (with and without byref), then loading, and getting the same results either way.  Also, testing that things like `a = Bar(); a.__class__ is Bar` work before and after the load_session, regardless of byref.  Things seem to work (eyeroll -- testing needed) for all supported versions of python.

mmckerns (approved):
LGTM.  Would be nice to have some tests, but can't complain as there are currently none.

leogama:
Added some tests. The Travis build is failing because of a warning raised in a assert statement before the proper tests. It's not critical...

leogama:
Travis doesn't like warnings. Can I substitute it by a print to stderr or a call to log()? Otherwise I will just comment out the offending lines.

mmckerns:
If the raising of a warning is what you want to test, then you can catch it in a try/except block, and assert True on except.  If this isn't the case, you can always use an if block to skip the line on travis, or something else.  What's the intent of the warning?

leogama:
I added the line to check if the modules that I choose to import in the tests weren't already loaded by python or dill. My assumption is that `import mod` may have different side effects than `from mod import obj` concerning the import of the module `mod`, its inclusion in `sys.modules` or any internal, implementation dependent, details. But I don't know if this precaution is necessary, and the list of default loaded modules varies from one python version to another... The actual tests are passing, this was just to catch an eventual false negative. I'm happy to suppress this line.

leogama:
Actually, I remove all the non "pre-loaded" modules from `sys.modules` before calling `load_session()`, so if one of the modules was already there from the beginning, there's no way to test if `load_session()` successfully imported it. And I thought that removing a module loaded by dill or python itself from `sys.modules` could cause problems, but maybe not.

mmckerns:
I would think that if you delete a module from `sys.modules` and from `__main__`, working in the test environment you control... then the module should not be available for `__main__` where you are testing until you `load_session`.

leogama:
I've removed the header "pre-test" and adapted the code. Travis builds are failing with Python 3.7 and Python 2.7. This last one "is allowed to fail", and all the tests pass with Python 2.7 on my machine though (Ubuntu Focal). The error is the same: the `xml.sax` module was not included in `sys.modules` after `load_session()`, but the `sax` module object was there. It occurred in the testing of `dump_session(byref=False)`, and didn't reach the `byref=True` case.

This is an artificial test setting however, if you start to think. The use case of `load_session()` is in a different session, not in the same one, and definitely not after tweaking `sys.modules` in this way. Maybe we should save the session to a file and test for correct module loading in a following, separate, test file?

mmckerns:
It might be worthwhile to try what you suggest.  That was my original use case, anyway.  FYI, a release is imminent (expected to be by end of day Friday or early Saturday, ET).  I know this is close to finished, so let me know if you need a bit more time and I can probably delay the release a bit if needed.

anivegesana:
Sorry for hijacking this PR, but @mmckerns I believe that this issue was actually present for a very long time. `recurse` broke all shared namespaces for children and it always has. This does seem like incorrect behavior, but it also seems like a new feature if it has been there for so long. It definitely would need more ironing out and we would have to push out the release a couple of days to fix it. Cloudpickle just uses [a weak ref dictionary](https://github.com/cloudpipe/cloudpickle/blob/f758eb34d1b3285dc582b73bfd8df4c47ed4fc68/cloudpickle/cloudpickle.py#L102) to keep track of the global dictionaries that it has created so far and the same pattern could be used here to solve the issue.

leogama:
> It might be worthwhile to try what you suggest. That was my original use case, anyway. FYI, a release is imminent (expected to be by midday Friday, ET). I know this is close to finished, so let me know if you need a bit more time and I will delay the release a bit if needed.

Please say you didn't close the release yet! I will finish this tests right now.

leogama:
@mmckerns How do I determine the order the tests are run? I've thought it was alphabetical order, but on my machine at least it seems it is not...

leogama:
A final strange error... Give me a couple minutes.

leogama:
Finally!!! 🚀 
All tests passed in all Python versions, from my side it's ready for merging. Thank you very much for your time and suggestions.

leogama:
Oh, sorry. 🤦🏼‍♂️ I thought I had to close the PR...

leogama:
> How do I determine the order the tests are run? I've thought it was alphabetical order, but on my machine at least it seems it is not...

The testing order is determined by the result of `glob.glob(...)` in `dill.tests.__main__.py`. The documentation says that whether the file names are sorted depends on the file system. Travis' file system seems to order them, but my machine's file system doesn't, therefore I wrapped the `glob()` call in a `sorted()` to guarantee the tests are run in lexical order. The locale is irrelevant as I numbered the test files for session saving/restoring.

mmckerns:
@leogama: I got stuck in a number of meetings today, and my whole day was pushed back, so I'm just getting to start with the releases now.  Which means, I'll probably review this tonight or early tomorrow, and the release will get built tomorrow.

tests/test_session_0.py
```diff
@@ -0,0 +1,185 @@
+#!/usr/bin/env python
+
+# Author: Leonardo Gama (@leogama)
+# Copyright (c) 2008-2016 California Institute of Technology.
```
mmckerns, line 4: nit: only need current year in copyright

tests/test_session_0.py
```diff
@@ -0,0 +1,185 @@
...
+        __main__.__dict__.update(copy_dict)
+        sys.modules.update(copy_modules)
+        del __main__, copy_dict, copy_modules, dump
+
+
+    # This is for code coverage, tests the use case of dump_session(byref=True)
+    # without imported objects in the namespace. It's a contrived example because
+    # even dill can't be in it.
+    from types import ModuleType
+    main = ModuleType('test_module')
```
mmckerns, line 141: need to clean up the files left on disk (i.e. *pyc or in the cache).  See `test_moduledict_where_not_main` in `test_nested.py`.

tests/test_session_0.py
```diff
@@ -0,0 +1,185 @@
...
+        test_file = dill._dill.StringIO(dump)
+        dill.load_session(test_file)
+    finally:
+        test_file.close()
+
+    assert x == 42
+
+
+    # Dump session for module that is not __main__:
+    import test_classdef as module
```
mmckerns, line 166: need to clean up the files left on disk (i.e. *pyc or in the cache).  See `test_moduledict_where_not_main` in `test_nested.py`.

tests/test_session_1_byref_false.py
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env python
```
mmckerns, line 1: filename is too long.  don't make it any longer than `test_session_1.py`

tests/test_session_2_byref_true.py
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env python
```
mmckerns, line 1: filename is too long.  don't make it any longer than `test_session_2.py`

tests/test_session_1_byref_false.py
```diff
@@ -0,0 +1,41 @@
...
+    except AssertionError:
+        import traceback
+        error_line = traceback.format_exc().splitlines()[-2].replace('[obj]', '['+repr(obj)+']')
+        print("Error while testing (byref=%s):" % byref, error_line, sep="\n", file=sys.stderr)
+        raise
+
+if __name__ == '__main__':
+    byref = False
+    dill.load_session(session_file % byref)
+    os.remove(session_file % byref)
```
mmckerns, line 38: should guard the `os.remove` in case the file is missing or does not exist

tests/test_session_0.py
```diff
@@ -0,0 +1,185 @@
...
+                    del __main__.__dict__[name]
+            for module in list(sys.modules.keys()):
+                if module not in original_modules:
+                    del sys.modules[module]
+
+            dill._test_file = dill._dill.StringIO(dump)
+            dill.load_session(dill._test_file)
+            #print(sorted(set(sys.modules.keys()) - original_modules))
+        finally:
+            dill._test_file.close()
```
mmckerns, line 129: should remove `*pkl` session files here as cleanup, if finishes correctly or errors out

tests/test_session_1_byref_false.py
```diff
@@ -0,0 +1,41 @@
...
+# Author: Leonardo Gama (@leogama)
+# Copyright (c) 2008-2016 California Institute of Technology.
+# Copyright (c) 2016-2022 The Uncertainty Quantification Foundation.
+# License: 3-clause BSD.  The full license text is available at:
+#  - https://github.com/uqfoundation/dill/blob/master/LICENSE
+
+from __future__ import print_function
+import dill, os, sys
+
+session_file = 'session-byref-%s.pkl'
```
mmckerns, line 12: this requires `test_session_0.py` to be run first.  instead you could import and generate the `*pkl` file if it doesn't exist.  I don't like the interdependency between these tests. Each test should be independently runnable, regardless of the order executed.

tests/test_session_2_byref_true.py
```diff
@@ -0,0 +1,27 @@
...
+# Copyright (c) 2016-2022 The Uncertainty Quantification Foundation.
+# License: 3-clause BSD.  The full license text is available at:
+#  - https://github.com/uqfoundation/dill/blob/master/LICENSE
+
+from test_session_1_byref_false import *
+
+if __name__ == '__main__':
+    byref = True
+    dill.load_session(session_file % byref)
+    os.remove(session_file % byref)
```
mmckerns, line 14: should guard the `os.remove` in case the file is missing or does not exist

mmckerns (changes requested):
I'm fine with the changes to `_dill.py` -- the tests need a few changes, however are generally good.  please address the comments left inline

tests/test_session_0.py
```diff
@@ -0,0 +1,185 @@
...
+original_objects = set(__main__.__dict__.keys())
+original_objects.add('original_objects')
+
+
+# Create various kinds of objects to test different internal logics.
+
+## Modules.
+import json                                         # top-level module
+import urllib as url                                # top-level module under alias
+import test_session_1_byref_false as local_mod      # non-builtin top-level module
```
mmckerns, line 23: need to clean up any files left on disk (i.e. *pyc or in the cache).

tests/test_session_0.py
leogama: Oops! I just copied it and didn't see. I usually set my own copyright notices in Vim and it puts the correct year for me... 😓

leogama: Should I put just this?
```
# Copyright (c) 2022 The Uncertainty Quantification Foundation.
```

tests/test_session_0.py
leogama: Do I have to clean up the files on each test file or can I clean up everything just in the last `test_session_*.py` (making sure it will do the clean up even if the run fails)?

tests/test_session_0.py
mmckerns: yep

tests/test_session_0.py
mmckerns: you should assume that everything is cleaned up after each test file is run

tests/test_session_0.py
leogama: These bytecode cache files are really annoying. What do you think of avoiding generating them by calling python in `dill.tests.__main__` with a `-B` flag or a `PYTHONDONTWRITEBYTECODE` env var? It only would be necessary to remove cache files at the end for `__init__.py` and `__main__.py` (which are still hanging there today, no?).

tests/test_session_1_byref_false.py
leogama: Agreed.

tests/test_session_1_byref_false.py
leogama: This won't work because for generating the `.pkl` file it will need to load all the modules necessary to populate the namespace. The idea behind saving on one file and loading in the next is to faithfully reproduce the use case of loading the dumped file in a fresh new session.

Also, _there are hidden spirits in the importing machinery seeing everything you do and playing tricks._ I tried to run all the tests in a single file at first but hit a weird bug where, after cleaning the `sys.modules` dictionary for the second time, the `calendar` module wouldn't show up there again, neither using the `__import__()` function nor the `import` statement and variants.

If you are convinced about having independent tests, the alternative I see is to use a python subprocess to test the file loading and module importing in a fresh interpreter session. Then the test files could be merged again.

tests/test_session_1_byref_false.py
leogama: Sure, pure distraction.

tests/test_session_2_byref_true.py
leogama: Idem.

tests/test_session_2_byref_true.py
leogama: Ibidem.

tests/test_session_0.py
leogama: This one was a mistake, it was supposed to be a dummy name for a dummy module. However, there is actually a file named `test_module.py` and it was accidentally imported... Solved the cleaning up problem with a function in `test_session_1.py`.

tests/test_session_0.py
leogama: Answering my second question, after doing the proper clean up in my test files, the cache files from `__init__.py` and `__main__.py` magically disappeared.

tests/test_session_1_byref_false.py
leogama: The subprocess solution worked flawlessly! It went out very clean and elegant in a single file (I was expecting to end with some macaroni code).

tests/test_session_0.py
leogama: solved

tests/test_session_0.py
leogama: solved

leogama:
Answered all the comments. All tests passing. Ready to go.

mmckerns (approved):
LGTM

leogama:
> @leogama: In case the OP doesn't open a `dill` GitHub issue, you may want to ponder this: https://stackoverflow.com/questions/72486294.

@mmckerns, take a look at my answer there. It is half a bug and half misdesign/misuse. It may need a decision about what to do.

mmckerns:
yep, that's my thought exactly

leogama:
I've already made up my mind. Restoring the whole namespace of a module to a different module (with a different name!) is a really bad idea, as it overwrites the second module's attributes like `__name__`, `__path__` and `__spec__` and creates orphaned classes and functions which dill itself can't deal with. I think `load_session()` should just keep the `main` parameter for backward compatibility and ignore it from now on.

The only doubt I have is whether `_import_module()` should support modules created at runtime if it fails to import. Should this work?:

```python
import dill, types
mod = types.ModuleType('runtime')
mod.var = 'spam'
dill.dump_session(main=mod)
```

In other session:
```python
import dill
mod = dill.load_session()
print(mod.var)
```

mmckerns:
I think I agree with you. I also think it'd be good if modules created at runtime (as in your example) worked.

mmckerns:
@leogama: ticket is opened here: #498

Issue #456, opened 2022-02-28
Dill Should Use a Custom dispatch_table vs Modifying pickle._Pickler.dispatch 


elmq0022:
Dill modifies python's pickler._Pickle.dispatch.  This means it can clash with similar packages with similar goals, notably cloudpickle.  

Dill does provide `dill.extend` to turn dill on and off, but this is cumbersome and relies on the user to be know when dill is or is not enabled.  

Using a custom [dispatch_table](https://docs.python.org/3/library/pickle.html#dispatch-tables) as specified in the Python's pickle documentation would be a cleaner implementation, in my opinion, as dill's extensions would be scoped only to dill and not clash with other packages polluting the global scope.

I haven't scoped the effort required, but I would be willing to assist if the maintainer(s) are willing to entertain such a change.

mmckerns:
`dill` made the choice to modify the contents of the `pickle` dispatch table, and unfortunately, `cloudpickle` made the same choice years later.  So, yes, this causes potential clashes.  One of the reasons this choice was made is historical, in that in python 2.x, `pickle` used a python dispatch table (and a pure python pickler) -- so, just by importing `dill`, one could augment the serialization of a package that is built to use `pickle`.  In python 3.x, pickle uses a C pickler, so the augment of the `_Pickle` dispatch table is less impactful.  Feel free to explore a PR along the lines you suggest.

PR #471, opened 2022-05-08
Add detailed trace mode showing saved object size and visual depth level

leogama:
This is a detailed mode for the trace feature to help to better visualize the hierarchy of the objects being dumped and also to identify large child objects. If it is OK, I will write the documentation.

Usage:

```python
import dill
dill.detect.trace(True, detail=True)
pickled = dill.dumps(obj)
```

### DIFF
--- a/dill/_dill.py
+++ b/dill/_dill.py
@@ -23,8 +23,9 @@ __all__ = ['dump','dumps','load','loads','dump_session','load_session',
 
 __module__ = 'dill'
 
-from .logger import adapter as logger
 import warnings
+from .logger import adapter as logger
+from .logger import trace as _trace
 
 import os
 import sys
--- a/dill/_dill.py
+++ b/dill/_dill.py
@@ -21,7 +21,7 @@ __all__ = ['dump','dumps','load','loads','dump_session','load_session',
            'UnpicklingError','HANDLE_FMODE','CONTENTS_FMODE','FILE_FMODE',
            'PickleError','PickleWarning','PicklingWarning','UnpicklingWarning']
 
-from .logger import adapter as logger, trace as _trace
+from .logger import adapter as logger
 import warnings
 
 import os
@@ -574,10 +574,10 @@ class Pickler(StockPickler):
         if NumpyUfuncType and numpyufunc(obj):
             @register(type(obj))
             def save_numpy_ufunc(pickler, obj):
-                _log_trace(pickler, "Nu: %s", obj)
+                logger.trace(pickler, "Nu: %s", obj)
                 name = getattr(obj, '__qualname__', getattr(obj, '__name__', None))
                 StockPickler.save_global(pickler, obj, name=name)
-                _log_trace(pickler, "# Nu")
+                logger.trace(pickler, "# Nu")
                 return
             # NOTE: the above 'save' performs like:
             #   import copy_reg
@@ -588,9 +588,9 @@ class Pickler(StockPickler):
         if NumpyDType and numpydtype(obj):
             @register(type(obj))
             def save_numpy_dtype(pickler, obj):
-                _log_trace(pickler, "Dt: %s", obj)
+                logger.trace(pickler, "Dt: %s", obj)
                 pickler.save_reduce(_create_dtypemeta, (obj.type,), obj=obj)
-                _log_trace(pickler, "# Dt")
+                logger.trace(pickler, "# Dt")
                 return
             # NOTE: the above 'save' performs like:
             #   import copy_reg
@@ -601,23 +601,18 @@ class Pickler(StockPickler):
         if NumpyArrayType and ndarraysubclassinstance(obj):
             @register(type(obj))
             def save_numpy_array(pickler, obj):
-                _log_trace(pickler, "Nu: (%s, %s)", obj.shape, obj.dtype)
+                logger.trace(pickler, "Nu: (%s, %s)", obj.shape, obj.dtype)
                 npdict = getattr(obj, '__dict__', None)
                 f, args, state = obj.__reduce__()
                 pickler.save_reduce(_create_array, (f,args,state,npdict), obj=obj)
-                _log_trace(pickler, "# Nu")
+                logger.trace(pickler, "# Nu")
                 return
         # end hack
         if GENERATOR_FAIL and type(obj) == GeneratorType:
             msg = "Can't pickle %s: attribute lookup builtins.generator failed" % GeneratorType
             raise PicklingError(msg)
-        else:
-            _log_trace(self)
-            try:
-                StockPickler.dump(self, obj)
-            finally:
-                _log_trace(self, reset=True)
-        return
+        logger.trace_setup(self)
+        StockPickler.dump(self, obj)
 
     dump.__doc__ = StockPickler.dump.__doc__
 
@@ -1164,9 +1159,9 @@ def _save_with_postproc(pickler, reduction, is_pickler_dill=None, obj=Getattr.NO
 
 #@register(CodeType)
 #def save_code(pickler, obj):
-#    _log_trace(pickler, "Co: %s", obj)
+#    logger.trace(pickler, "Co: %s", obj)
 #    pickler.save_reduce(_unmarshal, (marshal.dumps(obj),), obj=obj)
-#    _log_trace(pickler, "# Co")
+#    logger.trace(pickler, "# Co")
 #    return
 
 # The following function is based on 'save_codeobject' from 'cloudpickle'
@@ -1175,7 +1170,7 @@ def _save_with_postproc(pickler, reduction, is_pickler_dill=None, obj=Getattr.NO
 # License: https://github.com/cloudpipe/cloudpickle/blob/master/LICENSE
 @register(CodeType)
 def save_code(pickler, obj):
-    _log_trace(pickler, "Co: %s", obj)
+    logger.trace(pickler, "Co: %s", obj)
     if PY3:
         if hasattr(obj, "co_exceptiontable"):
             args = (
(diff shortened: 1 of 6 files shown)


dill/_dill.py
```diff
@@ -1176,7 +1211,7 @@ def _save_with_postproc(pickler, reduction, is_pickler_dill=None, obj=Getattr.NO
 # License: https://github.com/cloudpipe/cloudpickle/blob/master/LICENSE
 @register(CodeType)
 def save_code(pickler, obj):
-    log.info("Co: %s" % obj)
+    pickler._log_trace("Co: %s", obj)
```
anivegesana, line 1214: Because of limitations from long ago (#456), the pickler that is passed into the registered save functions is not necessarily a dill pickler and could have a missing `_log_trace` function.

I think that formatting the log trace as a tree is a good idea, but I think you should be using a `Formatter` instead of the pickler itself to do this:

https://stackoverflow.com/questions/35048921/format-log-messages-as-a-tree

leogama: Oh, now I get the whole `is_dill()` thing. It is related to other packages that also extend `pickle`. I'll take a look at `Formatter` then.

anivegesana: Other people can use the default pickler `pickle.Pickler` instead of `dill.Pickler` and they shouldn't see an error. This system is like this because in older versions of Python, the pickler class couldn't have its own dispatch. I think that it would be quick change to make the dispatch per class instead of global (cloudpickle did it), but it may be a breaking change and I didn't find time to verify that it isn't.

leogama: I think it's not so hard to verify whether the package is working after such a change, but for code that _extends_ dill it could be tricky. There are both benefits and drawbacks of being able to extend pickle just by importing dill. Maybe it could be made an optional feature somehow.

Hey, look at this! My notes from learning how to use `logging.Formatter` (it's a lot of fun):

```
Basic calling sequence with StreamHandler and Formatter in Python 3.7: 
(All methods from the second line on are from logging.__init__.py)     
                                                                       
dill.logging.logger.log ->   (or info, warning, etc.)                  
  Logger.log ->                                                        
    Logger._log ->                                                     
      Logger.makeRecord ->                                             
        LogRecord.__init__                                             
      Logger.handle ->                                                 
        Logger.callHandlers ->                                         
          Handler.handle ->                                            
            StreamHandler.emit ->                                      
              Handler.format ->                                        
                Formatter.format ->                                    
                  LogRecord.getMessage                                 
                  Formatter.formatMessage ->                           
                    PercentStyle.format                                
```

mmckerns:
So `dill` has a script that uses `objgraph` to get a visual understanding of some of the same information that `trace` provides.  Just FYI in case you aren't aware of it...

leogama:
> So `dill` has a script that uses `objgraph` to get a visual understanding of some of the same information that `trace` provides. Just FYI in case you aren't aware of it...

I did saw it. It's a good tool to get an overview of large objects or entire module structures. My proposed addition is meant for detailed inspection of a single or few objects in the pickle, and the "tree style" is just a visual hint to help you to find it in the middle of a long log. An option for `objgraph` to visualize object sizes would be interesting, though.

leogama:
Refactored the code using `logging` classes. Seems good.

mmckerns:
> I think that formatting the log trace as a tree is a good idea

I agree.

leogama:
Here is a sample output from the `test_dictviews.py` file:

```
> Dk: <dict_keys(['a'])>
>> Wr: <method 'keys' of 'dict' objects>
>>> F2: <function _getattr at 0x7fd372962820>
>>> # F2 (26 B)
>>> T1: <class 'dict'>
>>>> F2: <function _load_type at 0x7fd37295db80>
>>>> # F2 (17 B)
>>> # T1 (28 B)
>> # Wr (101 B)
> # Dk (156 B)
> Dv: <dict_values([1])>
>> Wr: <method 'values' of 'dict' objects>
>>> F2: <function _getattr at 0x7fd372962820>
>>> # F2 (26 B)
>>> T1: <class 'dict'>
>>>> F2: <function _load_type at 0x7fd37295db80>
>>>> # F2 (17 B)
>>> # T1 (28 B)
>> # Wr (105 B)
>> T1: <class 'enumerate'>
>> # T1 (18 B)
> # Dv (168 B)
> Di: <dict_items([('a', 1)])>
>> Wr: <method 'items' of 'dict' objects>
>>> F2: <function _getattr at 0x7fd372962820>
>>> # F2 (26 B)
>>> T1: <class 'dict'>
>>>> F2: <function _load_type at 0x7fd37295db80>
>>>> # F2 (17 B)
>>> # T1 (28 B)
>> # Wr (103 B)
>> D2: <dict object at 0x7fd37296ef00>
>> # D2 (9 B)
> # Di (116 B)
```

anivegesana:
Personally, I am more of a fan of blank space indentation or box drawing characters. If you think that is a good idea, I can look into later to see if it will look nicer that way without adding overhead.

https://en.m.wikipedia.org/wiki/Box_Drawing

anivegesana:
@leogama Thoughts?
```
┬ Dkvi: <dict_keys(['a'])>
├┬ Wr: <method 'keys' of 'dict' objects>
│├┬ F2: <function _getattr at 0x10af83b50>
││└ # F2 (26 B)
│├┬ T1: <class 'dict'>
││├┬ F2: <function _load_type at 0x10af82ef0>
│││└ # F2 (17 B)
││└ # T1 (28 B)
│└ # Wr (101 B)
├┬ D2: <dict object at 0x10ac88880>
│└ # D2 (9 B)
└ # Dkvi (114 B)
┬ Dkvi: <dict_values([1])>
├┬ Wr: <method 'values' of 'dict' objects>
│├┬ F2: <function _getattr at 0x10af83b50>
││└ # F2 (26 B)
│├┬ T1: <class 'dict'>
││├┬ F2: <function _load_type at 0x10af82ef0>
│││└ # F2 (17 B)
││└ # T1 (28 B)
│└ # Wr (105 B)
├┬ D2: <dict object at 0x10ac88880>
│└ # D2 (9 B)
└ # Dkvi (118 B)
┬ Dkvi: <dict_items([('a', 1)])>
├┬ Wr: <method 'items' of 'dict' objects>
│├┬ F2: <function _getattr at 0x10af83b50>
││└ # F2 (26 B)
│├┬ T1: <class 'dict'>
││├┬ F2: <function _load_type at 0x10af82ef0>
│││└ # F2 (17 B)
││└ # T1 (28 B)
│└ # Wr (103 B)
├┬ D2: <dict object at 0x10ac88880>
│└ # D2 (9 B)
└ # Dkvi (116 B)
┬ Dkvi: <dict_values([1])>
├┬ Wr: <method 'values' of 'dict' objects>
│├┬ F2: <function _getattr at 0x10af83b50>
││└ # F2 (26 B)
│├┬ T1: <class 'dict'>
││├┬ F2: <function _load_type at 0x10af82ef0>
│││└ # F2 (17 B)
││└ # T1 (28 B)
│└ # Wr (105 B)
├┬ D2: <dict object at 0x10ac88880>
│└ # D2 (9 B)
└ # Dkvi (118 B)
┬ Dkvi: <dict_items([('a', 1)])>
├┬ Wr: <method 'items' of 'dict' objects>
│└ # Wr (53 B)
└ # Dkvi (59 B)
┬ Dkvi: <dict_keys(['a'])>
├┬ Wr: <method 'keys' of 'dict' objects>
│└ # Wr (51 B)
└ # Dkvi (57 B)
```

leogama:
Sleek! I will just add a fallback clause for non-UTF-8 terminals just in case.

anivegesana:
Ah yes. Keep forgetting that Python 2 is still supported by dill. Glad you like it!

leogama:
> Ah yes. Keep forgetting that Python 2 is still supported by dill.

Actually, I had to work really hard to get ASCII output, which I eventually did with `env -i LC_ALL=C python2`. It seems that Python 2 also uses UTF-8 encoding for terminal output in most cases. But there must be some really dumb terminals out there.

And the Unicode strings were necessary just for the `translate()` method to work with Unicode characters. Otherwise it was working just fine.

@mmckerns Any final thoughts?

dill/logger.py
```diff
@@ -0,0 +1,153 @@
...
+# -*- coding: utf-8 -*-
+#
+# Author: Leonardo Gama (@leogama)
+# Copyright (c) 2022 The Uncertainty Quantification Foundation.
+# License: 3-clause BSD.  The full license text is available at:
+#  - https://github.com/uqfoundation/dill/blob/master/LICENSE
+r"""
+Logging utilities for dill.
+
+Python's logging functionality can be conceptually divided into five steps:
```
mmckerns, line 11: What's the point of this documentation? (From here to "NOTE")
I'd expect the module docs to cover what logging customization was done, but not cover basic logging. So, what you have under the "customization" header, plus maybe some detail on INFO_DETAIL.  Possibly also any expected formatting for entries to the new `log_trace` method... and an example and any explanation of the details of the printed logger trace that you want to discuss.

dill/logger.py
```diff
@@ -0,0 +1,153 @@
...
+adapter = TraceAdapter(logger)
+
+handler = logging.StreamHandler()
+logger.addHandler(handler)
+
+formatter = TraceFormatter("%(prefix)s%(message)s%(suffix)s")
+handler.setFormatter(formatter)
+
+
+def trace(boolean, detail=False):
```
mmckerns, line 148: Why bother to have the option to turn off a more detailed trace?  Is there any value to a less detailed trace?  If not, then don't expose an option to turn off `detail`.  The new trace is nice... why wouldn't someone always use it?

mmckerns (changes requested):
Please address the comments.  Otherwise, LGTM.

dill/logger.py
leogama: I agree with you that this documentation of Python's logging framework doesn't fit in the module's documentation. However, I don't think this is "basic logging", but rather "basic logging development". I couldn't find this information in the official documentation, nor in the first resources about logging I've found online, that mostly focused in basic logging usage. I had to untangle the library code to understand that, to do what I wanted to do, I needed to create a `LoggerAdapter` to wrap the `Logger` returned by `logging.getLogger('dill')` (taking care to handle the `process()` method). I just thought that this info should be kept with the implementation. What if this documentation chunk is moved to a comment just above the classes definition? I don't remember if I wrote down this rationale of the classes design, but if not I can add it.

Will complement the remaining as you suggested. 👍🏼

mmckerns: > What if this documentation chunk is moved to a comment just above the classes definition?

Perfect

dill/logger.py
leogama: Sorry, I can't recall why I thought this should be optional. The only scenario I can think of in which it would be useful to not indent the trace in a tree style is for inspecting the trace of some (very) deeply nested object structure. But even then I don't think the indentation would obscure it much... Maybe it's better to just drop the option (and the `INFO_DETAIL` log level with it).

mmckerns: I don't see any reason for switching off `detail`, so if you don't either, then remove the option.

leogama: Option removed.

dill/_dill.py
```diff
@@ -23,14 +23,7 @@
...
 
-import logging
-log = logging.getLogger("dill")
-log.addHandler(logging.StreamHandler())
-def _trace(boolean):
-    """print a trace through the stack when pickling; useful for debugging"""
-    if boolean: log.setLevel(logging.INFO)
-    else: log.setLevel(logging.WARN)
-    return
+from .logger import adapter as logger
```
mmckerns, line 26: How about we add: `from .logger import trace as _trace` to minimize existing code breakage due to this PR?  Otherwise, LGTM.

mmckerns (changes requested):
Let's make the suggested change to improve code compatibility across this PR.

dill/_dill.py
mmckerns: People should be using `dill.detect.trace`.  However... I've found that if it's easy to keep backward compatibility, you should do it.  Here the cost essentially nothing because we are already importing from the dill logging module.

leogama:
@mmckerns and @anivegesana: what do you think of converting `trace` to a context manager?

```python
class trace:
    """print a trace through the stack when pickling; useful for debugging"""
    def __init__(self, boolean=None):
        self.boolean = boolean
        if boolean is not None:
            logger.setLevel(logging.INFO if boolean else logging.WARNING)
    def __enter__(self):
        self.previous_level = logger.level
        logger.setLevel(logging.INFO)
        return logger
    def __exit__(self, *exc_info):
        logger.setLevel(self.previous_level)
        return False
    def __repr__(self):
        if self.boolean is None:
            return super().__repr__()
        return "dill trace is %s" % ("ON" if self.boolean else "OFF")
```

It can be used in the same way as before too, but the using it as a context manager preserves the previous log level:
```python
>>> from dill import detect, dumps
>>> from dill.logger import logger
>>> d = {'a': 1, 'b': {}}
>>> logger.setLevel(40)
>>> detect.trace(True)
dill trace is ON
>>> _ = dumps(d)
┬ D2: <dict object at 0x7f83c2176180>
├┬ D2: <dict object at 0x7f83c2188c40>
│└ # D2 [2 B]
└ # D2 [16 B]
>>> detect.trace(False)
dill trace is OFF
>>> _ = dumps(d)  # no trace
>>> logger.level
30
>>> logger.setLevel(40)
>>> with detect.trace():
>>>     dumps(d)
┬ D2: <dict object at 0x7f83c2176180>
├┬ D2: <dict object at 0x7f83c2188c40>
│└ # D2 [2 B]
└ # D2 [16 B]
>>> _ = dumps(d)  # no trace
>>> logger.level
40
```

mmckerns (commented):
LGTM.  Will you be adding the context manager to this PR?  I'll approve or defer depending on your response.

leogama:
> LGTM. Will you be adding the context manager to this PR? I'll approve or defer depending on your response.

Yes, I prefer to finish this in one shot. Let me just update the documentation and add a test and we can close it.

leogama:
I've made the context manager implementation separated lately.

dill/logger.py
```diff
@@ -0,0 +1,276 @@
...
+    Arguments:
+        arg: a boolean value or an optional ``PathLike`` object for the context manager
+        mode: mode string for ``open()`` if a file name is passed as the first argument
+    """
+    if not isinstance(arg, bool):
+        return _trace(filename=arg, mode=mode)
+    logger.setLevel(logging.INFO if arg else logging.WARNING)
+
+@contextmanager
+def _trace(filename, mode):
```
mmckerns, line 259: I think it is more robust if it takes a file handle, instead of a filename.  You could then do like pandas... and open the file if given the filename.

mmckerns: Are you ok with the behavior that happens if someone gives `dill.detect.trace('test.log')` without using a context?

mmckerns (changes requested):
Please address the two comments on the context manager.  Otherwise, LGTM.

dill/logger.py
leogama: _First comment:_ Taking an open stream is good if the user already have an open file, otherwise they would need to use two nested `with` statements or a `with` statement with two context managers (which I don't know if most users are aware of) like:
```python
with open('/tmp/trace', 'w') as file, dill.detect.trace(file) as log:
    # code
```

I think it's better to cover both use cases then.

---

_Second comment:_ No, not really... The first, class based, implementation avoided side effects if not called from a `with` statement. I'll revert to that.

mmckerns (approved):

PR #472, opened 2022-05-08
Incidental implementation specific types

anivegesana:
https://github.com/uqfoundation/dill/pull/449#issuecomment-1103998082

Extended the list of inaccessible Python types to include types that were added as recently as Python 3.11.

### DIFF
--- a/dill/_dill.py
+++ b/dill/_dill.py
@@ -40,6 +40,7 @@ PY3 = (sys.hexversion >= 0x3000000)
 OLDER = (PY3 and sys.hexversion < 0x3040000) or (sys.hexversion < 0x2070ab1)
 OLD33 = (sys.hexversion < 0x3030000)
 OLD37 = (sys.hexversion < 0x3070000)
+OLD38 = (sys.hexversion < 0x3080000)
 OLD39 = (sys.hexversion < 0x3090000)
 OLD310 = (sys.hexversion < 0x30a0000)
 PY34 = (0x3040000 <= sys.hexversion < 0x3050000)
@@ -85,6 +86,7 @@ import marshal
 import gc
 # import zlib
 from weakref import ReferenceType, ProxyType, CallableProxyType
+from collections import OrderedDict
 from functools import partial
 from operator import itemgetter, attrgetter
 # new in python3.3
@@ -271,8 +273,6 @@ except NameError:
     try: ExitType = type(exit) # apparently 'exit' can be removed
     except NameError: ExitType = None
     singletontypes = []
-
-from collections import OrderedDict
 
 import inspect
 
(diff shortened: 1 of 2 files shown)


dill/_dill.py
```diff
@@ -730,18 +738,89 @@ def _create_typemap():
     'PyBufferedReaderType': PyBufferedReaderType,
     'PyBufferedWriterType': PyBufferedWriterType,
     'PyTextWrapperType': PyTextWrapperType,
-})
+
+    "BytesIteratorType": type(iter(b'')),
```
mmckerns, line 742: `BytesIteratorType` is not found in` dill._objects`.

mmckerns (commented):
All of the objects currently in `_reverse_typemap` are also later fed to `register` (with the exception of the `ExitType`).  I believe that's primarily why some objects were in the `_reverse_typemap`, and others were not.  You've added a number of objects... and I assume none have been registered because pickle will handle those objects without intervention from `dill`.  If that's so, then what's the point of adding them to the `_reverse_typemap`?  If it's just to have them included in `dill`, so one can grab an object of that type if needed... then I'd argue they could just be added to `dill._objects` and left out of `_reverse_typemap`. Otherwise, LGTM.

anivegesana:
I am still on the fence about including the types themselves in the reverse typemap. I think it makes perfect sense to put them in objects.py because the objects are commonly used by Python users, but the types themselves are more of an internal detail that I don't think anyone uses directly. PyPy and CPython iterate over builtin collections differently leading to differences there which make it a little bit more confusing.

anivegesana:
That definition is a little overloaded. `XRangeIteratorType` objects can be pickled but the type `XRangeIteratorType` itself cannot. Don't think that anyone should pickle it, but it would make `_reverse_typemap` incomplete. I'll remove it and if anyone needs it they can open an issue and add it later.

mmckerns:
Let me know when you want me to review this again.

anivegesana:
I think it is good now. Don't know why the coverage went down when I removed code. I will see if it is something that I did.

dill/_dill.py
```diff
@@ -732,18 +740,62 @@ def _create_typemap():
...
+    _incedental_reverse_typemap['WrapperDescriptorType'] = WrapperDescriptorType
+    _incedental_reverse_typemap['MethodDescriptorType'] = MethodDescriptorType
+    _incedental_reverse_typemap['ClassMethodDescriptorType'] = ClassMethodDescriptorType
 else:
-    _reverse_typemap['MemberDescriptorType'] = MemberDescriptorType
+    _incedental_reverse_typemap['MemberDescriptorType'] = MemberDescriptorType
+
+try:
+    import symtable
+    _incedental_reverse_typemap["SymtableStentryType"] = type(symtable.symtable("", "string", "exec")._table)
```
mmckerns, line 776: Why `SymtableStentryType` as opposed to `SymtableEntryType`?

anivegesana:
I called it `SymtableStentryType` because that is what the name of the type is (before changing from snake case to camel case.) I will change it to `SymtableEntryType` to be a little more consistent.

mmckerns:
The few places I found references to it it was as `SymtableEntryType`, so I prefer that. Thanks.

PR #473, opened 2022-05-10
Create a soft_def shim to combine move_to and Getattr

### LINKED ISSUE #447: single `shims` function calls `move_to` and `Getattr`
The `@move_to` decorator must be paired with the overloading by the `Getattr` later in the `shims` module.  It seems like with a little redesign, this could be consolidated to a single function / decorator.

anivegesana:
New (hopefully more intuitive) construct to combine move_to and Getattr into one function call.

Resolves #447

### DIFF
--- a/dill/_shims.py
+++ b/dill/_shims.py
@@ -150,11 +150,43 @@ def move_to(module, name=None):
         return func
     return decorator
 
+def soft_def(name, default):
+    """
+    A easier to understand and more compact way of "softly" defining a function.
+    These two pieces of code are equivalent:
+
+    if _dill.OLD3X:
+        def _create_class():
+            ...
+    _create_class = soft_def('_create_class', types.new_class)
+
+    if _dill.OLD3X:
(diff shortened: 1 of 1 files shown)


mmckerns (approved):
This looks good, in general. I don't like the name `soft_def`.  You aren't actually "defining" the shim, you kind of do that above when you write the function code... it's more like we `register` the shim in `_dill`.  @anivegesana: What do you think of the name `register`, or something similar?  Otherwise LGTM, and will pull when we agree on a name.

anivegesana:
`register` is already taken by the register function for the pickler. What about `if_not_defined_in` or `ifndef_in`?

mmckerns:
How about: `register_shim`?  Isn't that what you are doing?  You are registering the shim in `_dill` (and setting a default).  Also... one other thought... is it worthwhile to enable passing the module as a keyword arg (with `_dill` as the default)?

PR #477, opened 2022-05-17
Support PyCapsule

### LINKED ISSUE #354: Can PyCapsule be pickled?
Easy way to produce a PyCapsule object:
```
Python 3.6.10 (default, Dec 21 2019, 11:39:07) 
[GCC 4.2.1 Compatible Apple LLVM 9.0.0 (clang-900.0.39.2)] on darwin
>>> import dill
>>> import ctypes
>>> pycapsule = ctypes.pythonapi.PyCapsule_New
>>> pycapsule.restype = ctypes.py_object
>>> c = pycapsule()
>>> c   
\ufffd" at 0x1040f2330>d\ufffdtcH\ufffd
>>> type(c)
<class 'PyCapsule'>
>>> dill.dumps(c)
  File "/Users/mmckerns/lib/python3.6/site-packages/dill-0.3.2.dev0-py3.6.egg/dill/_dill.py", line 265, in dumps
    dump(obj, file, protocol, byref, fmode, recurse, **kwds)#, strictio)
  File "/Users/mmckerns/lib/python3.6/site-packages/dill-0.3.2.dev0-py3.6.egg/dill/_dill.py", line 259, in dump
    Pickler(file, protocol, **_kwds).dump(obj)
  File "/Users/mmckerns/lib/python3.6/site-packages/dill-0.3.2.dev0-py3.6.egg/dill/_dill.py", line 445, in dump
    StockPickler.dump(self, obj)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/pickle.py", line 409, in dump
    self.save(obj)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/pickle.py", line 496, in save
    rv = reduce(self.proto)
TypeError: can't pickle PyCapsule objects
```

Interface is here: https://github.com/python/cpython/blob/master/Include/pycapsule.h

anivegesana:
Fixes #354

### DIFF
--- a/dill/_dill.py
+++ b/dill/_dill.py
@@ -1059,11 +1059,50 @@ else:
         t = collections.namedtuple(name, fieldnames, defaults=defaults, module=modulename)
         return t
 
+# TODO: Remove this copy when #450 is pulled
+# https://github.com/python/cpython/blob/a8912a0f8d9eba6d502c37d522221f9933e976db/Lib/pickle.py#L322-L333
+def _getattribute(obj, name):
+    for subpath in name.split('.'):
+        if subpath == '<locals>':
+            raise AttributeError("Can't get local attribute {!r} on {!r}"
+                                 .format(name, obj))
+        try:
+            parent = obj
+            obj = getattr(obj, subpath)
+        except AttributeError:
+            raise AttributeError("Can't get attribute {!r} on {!r}"
+                                 .format(name, obj)) # In Py3: from None
+    return obj,
+
 def _create_capsule(pointer, name, context, destructor):
+    attr_found = False
     try:
-        # TODO: Somehow check this condition?
-        return _PyCapsule_Import(name, False)
+        # based on https://github.com/python/cpython/blob/f4095e53ab708d95e019c909d5928502775ba68f/Objects/capsule.c#L209-L231
+        if PY3:
+            uname = name.decode('utf8')
+        else:
+            uname = name
+        for i in range(1, uname.count('.')+1):
+            names = uname.rsplit('.', i)
+            try:
+                module = __import__(names[0])
+            except:
+                pass
+            obj = module
+            for attr in names[1:]:
+                obj = getattr(obj, attr)
+            capsule = obj
+            attr_found = True
+            break
     except:
+        pass
+
+    if attr_found:
+        if _PyCapsule_IsValid(capsule, name):
+            return capsule
+        raise UnpicklingError("%s object exists at %s but a PyCapsule object was expected." % (type(capsule), name))
+    else:
+        warnings.warn('Creating a new PyCapsule %s for a C data structure that may not be present in memory. Segmentation faults or other memory errors are possible.' % (name,), UnpicklingWarning)
         capsule = _PyCapsule_New(pointer, name, destructor)
         _PyCapsule_SetContext(capsule, context)
         return capsule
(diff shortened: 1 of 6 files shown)


dill/_dill.py
```diff
@@ -1059,6 +1059,54 @@ def _create_namedtuple(name, fieldnames, modulename, defaults=None):
         t = collections.namedtuple(name, fieldnames, defaults=defaults, module=modulename)
         return t
 
+# TODO: Remove this copy when #450 is pulled
```
mmckerns, line 1062: This is duplicate code and can be removed.

tests/test_objects.py
```diff
@@ -57,6 +59,36 @@ def test_objects():
...
+                    warnings.simplefilter("ignore", pickle.PicklingWarning)
+                    pickle.copy(capsule)
+            except pickle.UnpicklingError:
+                pass
+            else:
+                raise AssertionError("Expected a different error")
 
 if __name__ == '__main__':
     test_objects()
+    if test_pycapsule is not None:
```
mmckerns, line 93: this test should be moved elsewhere. Its own test module is fine.  Note that `PyCapsule` objects are not yet in `dill._objects`.

dill/_dill.py
```diff
@@ -1059,6 +1059,54 @@ def _create_namedtuple(name, fieldnames, modulename, defaults=None):
...
+            break
+    except:
+        pass
+
+    if attr_found:
+        if _PyCapsule_IsValid(capsule, name):
+            return capsule
+        raise UnpicklingError("%s object exists at %s but a PyCapsule object was expected." % (type(capsule), name))
+    else:
+        warnings.warn('Creating a new PyCapsule %s for a C data structure that may not be present in memory. Segmentation faults or other memory errors are possible.' % (name,), UnpicklingWarning)
```
mmckerns, line 1105: I'm on the fence about the use of a warning. You also added warnings in a previous PR related to pickling `numpy` objects. Warnings can be annoying, especially when you are running the code in parallel and they can be triggered on each new process.

mmckerns (commented):
Minor adjustments needed here, otherwise LGTM.

tests/test_objects.py
anivegesana: Thanks for the reminder.

dill/_dill.py
mmckerns: That's true. Maybe it would be nice to have `dill.settings['warnings']` which does exactly what you show above.  What do you think of that?

anivegesana: I do not know whether it would be useful to do it globally or per pickler. Using the warnings package does make it a little bit more flexible allowing users to turn the warnings into errors, ignore all but the first time that the warning appears as well as filter some dill warnings but not others.

We could add the `'warnings'` setting and it would just change the filter to `"ignore"` if False or `"default"` if True as a shortcut while allowing others to use the functionality in the warnings package.

tests/test_pycapsule.py
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env python
+#
+# Author: Mike McKerns (mmckerns @caltech and @uqfoundation)
```
mmckerns, line 3: you can put yourself as the author, and trim the copyright to only this year.

mmckerns (approved):
LGTM.  I'll give you some time to change the Author and Copyright on the new test file.  I'll pull the PR in the morning.

dill/_dill.py
mmckerns: let's not worry about it now.  But something to think about

PR #492, opened 2022-05-31
tests: fix a test for session saving

leogama:
The created module `__test_main__` was being loaded in `__main__`.


mmckerns (approved):
LGTM

PR #475, opened 2022-05-11
Add options to exclude objects from dump_session()

leogama:
#### `__init__.py`

- Rolled back bad reordering of name imported from `._dill`
- Added the function `is_pickled_module()` to `dill`s namespace
- Renamed imported submodule `logger` to `logging`
- Moved `import reload` to inside the function where it's used to clear it from the global namespace

#### `_dill.py`

- Added a paragraph to the module's docstring pointing to the main functions (with links in the Sphinx generated document)
- Added `from __future__ import annotations`
- Use the submodule `dill.logging` in place of the stdlib `logging` module due to the new `TRACE` logging level
- Rolled back the reordering of names as in `__init__.py`
- Moved the import of `dataclasses` up, close to other imports, and moved the definition of `EXTENSION_SUFFIXES` down, close to a new `PYTHONPATH_PREFIXES` constant and the definition of `_is_builtin_module()`, which uses both.
- Renamed `singletontypes` to `IPYTHON_SINGLETONS`, a more descriptive name
- Joined some exception types from `pickles()` into a new constant `UNPICKLEABLE_ERRORS`, which is also used elsewhere for `refonfail`
- Created a helper function `_getopt()` to get an option from a named argument, falling back to settings, to reduce the clutter
- `Pickler`:
    - Added a description to the `dispatch` attribute and a Sphinx directive to hide the giant and uninformative representation of the dictionary that was included in the documentation page
    - Added flags `_refimported`, `_refonfail` and `_first_pass` with default (`False`) values to the `Pickler` class (should this remove the need to test for `pickler._session` before testing for them? not related to `is_dill(pickler)` tests)
    - Substituted the `_file` attribute of `Pickler` instances by a more specific `_file_tell`, used by the logger to compute the partial pickle sizes —following the `self.file_write` pattern from the `pickle` module (should this also be public?)
- `Pickler.save()`:
    - No need to test for the truthiness of `Numpy*Type` in each `if` test with a previous test of `NumpyArrayType`
    - Implementation of rollback if object fails to pickle with `refonfail`
    - An unpickleable variable in `__main__` or in a module-type object raises an exception
    - An unpickleable variable in another module makes the whole module to be saved by reference with a warning
- Substitute `bytes('0', 'UTF-8')` instances by the `pickle.POP` constant
- Moved `_module_map()` and `_lookup_module()` back here from `session.py`, as they are also used by the `refonfail` feature
[...]

PR #499, opened 2022-06-04
Kickstart support drop for Python < 3.7

leogama:
@mmckerns and @anivegesana, plus anyone interested: With the prospect of dill 0.3.6 release —it would be fun if it matched the lowest Python version supported, 3.7— in the near future, it is an opportunity to remove some cruft from the code base after the official support drop for Python 2 and Python < 3.7. Important: _this is not a refactoring_.

I propose to do so in discrete steps to facilitate review and minimize the chances of introducing bugs. The first steps should only remove code that Python/PyPy >= 3.7 will never run, with minimal code rewrite beyond necessary indentation changes and substitution of variables left as unconditional constants, after the other changes, by its literal values.

I plan to do each step in a separated commit, so that you guys can review the individual diffs and any unit test failure is easy to catch:

- [x] 1. Remove unreachable code and unused Python version constants (e.g. `IS_PYPY2`), and convert the new unconditional constants left to literals.
- [x] 2. Convert conditional code written as strings passed `exec()` due to syntax limitations to common if-else code. Remove conditional imports. Remove `from __future__` imports.
3. Clean-up now unused types definitions if they are not in `_reverse_typemap`, to still allow unpickling of objects created with older combinations of python+dill versions. Unalias some types for internal usage, like BytesIO that is renamed to StringIO —it was a bit confusing the first time I met it. → **moved to** #504
4. Rewrite some snippets of code to take advantage of new Python 3 syntax if it improves code quality or speed, or if the used functions are deprecated in 3.7 → **moved to** #505

---

Neither all steps are strictly necessary, nor they need to be completed before the next release. **If you have the time, it would be important to thoroughly review every change made (the more eyeballs, the better).** Please feel completely free to make changes by committing directly here and to add or remove steps to the ones I'm suggesting.


anivegesana:
There are a couple more that are no longer needed. `save_slice` is one of them, but it will require a little bit of reasoning and trying them all out in pickle to see which ones are in CPython now.

leogama:
> There are a couple more that are no longer needed. `save_slice` is one of them, but it will require a little bit of reasoning and trying them all out in pickle to see which ones are in CPython now.

Yes, the now obsolete `save_<type>()` functions can be safely removed in my proposed step 3, while keeping they correspondent `_create_<type>()` functions for backward compatibility. And you are right in that every component removed and modified in this critical step must be evaluated carefully and tested.

mmckerns:
I'd split your steps into three different PRs -- (1 & 2), (3), and (4).  Official end of support for 3.7 from python.org is 2023-06-27.  So that means `dill` will end support for 3.7 in 2023 or 2024.  It's still a good way off.

leogama:
> I'd split your steps into two or three different PRs -- keep this one as (1 & 2), and create a new PR for (4). I'm not sure the best place for (3).

Agreed. Is there a way to create an empty draft PR? Or do I nead at least a stub change?

leogama:
@mmckerns I managed to open the PRs with empty commits. You can add the milestone to this one now.

dill/__diff.py
```diff
@@ -10,18 +10,12 @@
 Module to show if an object has changed since it was memorised
 """
 
-import os
-import sys
-import types
+import builtins, os, sys, types
```
mmckerns, line 13: don't condense module imports to a single line

leogama: Just googled about this. The "one module import by line" style is the PEP8 recommendation (although it doesn't give a rationale) and it seems to avoid merge conflicts in some cases, which is nice. Okay

tests/__main__.py
```diff
@@ -5,15 +5,12 @@
 # License: 3-clause BSD.  The full license text is available at:
 #  - https://github.com/uqfoundation/dill/blob/master/LICENSE
 
-from __future__ import print_function
-import glob
-import os
+import glob, os, subprocess as sp
```
mmckerns, line 8: don't do multiple module-level imports on a single line

tests/__main__.py
```diff
@@ -5,15 +5,12 @@
...
 try:
     import pox
-    python = pox.which_python(version=True, fullpath=False) or 'python'
+    python = pox.which_python(version=True, fullpath=False) or 'python3'
```
mmckerns, line 11: We have to look this up (it's in one of the UQFoundation's tickets)... but one of (`python`, `python3`) is an alias that isn't guaranteed to be installed on all systems.  So, this change may be fine, and maybe not.

leogama: This got me a couple times when testing on different Python versions, and it run Python 2 instead (at least I noticed something was wrong, though.

There's a function `shutil.which()` in since Py3.3 that could be used to determine if there's a `python3` command and fall back to `python` as a default. Something like:
```python
default_python = 'python3' if shutil.which('python3') is not None else 'python'
```

mmckerns: Here's what I was referring to: https://github.com/uqfoundation/pox/pull/31.  Some environments don't have a `python3`.

dill/detect.py
```diff
@@ -87,12 +75,8 @@ def code(func):
...
     '''
-    if PY3:
-        im_func = '__func__'
-        func_code = '__code__'
-    else:
-        im_func = 'im_func'
-        func_code = 'func_code'
+    im_func = '__func__'
+    func_code = '__code__'
     if ismethod(func): func = getattr(func, im_func)
```
mmckerns, line 80: I expect these lines could be condensed, and getattr no longer needed.

dill/detect.py
```diff
@@ -143,14 +123,9 @@ def freevars(func):
...
-        func_code = '__code__'
-        func_closure = '__closure__'
-    else:
-        im_func = 'im_func'
-        func_code = 'func_code'
-        func_closure = 'func_closure'
+    im_func = '__func__'
+    func_code = '__code__'
+    func_closure = '__closure__'
     if ismethod(func): func = getattr(func, im_func)
```
mmckerns, line 129: I expect these lines could be condensed, and getattr no longer needed.

dill/detect.py
```diff
@@ -199,16 +174,10 @@ def globalvars(func, recurse=True, builtin=False):
...
-    else:
-        im_func = 'im_func'
-        func_code = 'func_code'
-        func_globals = 'func_globals'
-        func_closure = 'func_closure'
+    im_func = '__func__'
+    func_code = '__code__'
+    func_globals = '__globals__'
+    func_closure = '__closure__'
     if ismethod(func): func = getattr(func, im_func)
```
mmckerns, line 181: I expect these lines could be condensed, and getattr no longer needed.

tests/__main__.py
leogama: Are you sure? He says Fedora Linux doesn't have a `python` executable, only `python3`. My suggestion above is almost identical to yours in that issue by the way...

mmckerns:
I've reviewed all files except `_dill.py` and `_objects.py`, and all looks good thus far. I've commented where there are/were issues. Will finish this tomorrow.

tests/__main__.py
mmckerns: whoops. I misread that. :)

dill/_dill.py
```diff
@@ -437,8 +373,7 @@ def _stash_modules(main_module):
     imported_as = []
     imported_top_level = []  # keep separeted for backwards compatibility
     original = {}
-    items = 'items' if PY3 else 'iteritems'
-    for name, obj in getattr(main_module.__dict__, items)():
+    for name, obj in vars(main_module).items():
```
mmckerns, line 376: Not sure if this should stay `main_module.__dict__` as opposed to `vars(main_module)`.  They are not always the same.

leogama: According to [documentation](https://docs.python.org/3/library/functions.html#vars) (including from Py2.7), it _is_ always the case:

> `vars([object])`
> Return the `__dict__` attribute for a module, class, instance, or any other object with a `__dict__` attribute.
> (...)
> A `TypeError` exception is raised if an object is specified but it doesn’t have a `__dict__` attribute (...).

This is good if we have to distinguish a missing `__dict__` from an `AttributeError` in the same code block. However, just above the `vars()` documentation is `type()` documentation, and it states:

> `class type(object)`
> (...)
> With one argument, return the type of an object. The return value is a type object and **generally** the same object as returned by `object.__class__`.

Although I don't know what _that_ means...

leogama: By the way, I think I changed only this instance to match what I used in other session functions.
 
Regardless of the type of exception they raise on failure, `obj.__dict__` and `vars(obj)` can be considered just different styles. Both are used extensively in the code base: `_dill` and `_objects` tend to use the former, while the modules `detect` and `source` (and the new `session` module in #475 ) use mostly the latter. It would be good to standardize this.

I like the fact that `vars(module)` is close in syntax to `globals()`, and also the distinct error type 
—that I wasn't aware of actually.

dill/_dill.py
```diff
@@ -980,24 +900,6 @@ def _create_code(*args):
             return CodeType(args[0], *argz)
         elif len(args) == 15: return CodeType(*args)
         return CodeType(args[0], 0, *args[1:]) # from 2.7
-    # obj.co_argcount, obj.co_nlocals, obj.co_stacksize, obj.co_flags,
```
mmckerns, line 983: We should either outdent the last `elif` (for python 3.7), or convert it to an `else`.

dill/_dill.py
```diff
@@ -1037,12 +939,9 @@ def _create_filehandle(name, mode, position, closed, open, strictio, fmode, fdat
         import tempfile
         f = tempfile.TemporaryFile(mode)
     else:
-        # treat x mode as w mode
```
mmckerns, line 1040: I think this comment should not be removed

dill/_dill.py
```diff
@@ -1281,14 +1134,14 @@ def _getattr(objclass, name, repr_str):
...
-    except:
+    except Exception:
         try:
             attr = objclass.__dict__
             if type(attr) is DictProxyType:
                 attr = attr[name]
             else:
                 attr = getattr(objclass,name)
-        except:
+        except (AttributeError, KeyError):
```
mmckerns, line 1144: ...I think this is correct.

dill/_dill.py
```diff
@@ -1829,7 +1595,7 @@ def save_wrapper_descriptor(pickler, obj):
 def save_cell(pickler, obj):
     try:
         f = obj.cell_contents
-    except:
+    except ValueError:
```
mmckerns, line 1598: Not sure if this is always a VaueError...

dill/_dill.py
```diff
@@ -2117,24 +1868,21 @@ def save_property(pickler, obj):
 @register(classmethod)
 def save_classmethod(pickler, obj):
     log.info("Cm: %s" % obj)
-    im_func = '__func__' if PY3 else 'im_func'
+    im_func = '__func__'
```
mmckerns, line 1871: could just use `__func__` in `getattr` directly below.

dill/_dill.py
```diff
@@ -2117,24 +1868,21 @@ def save_property(pickler, obj):
...
     try:
         orig_func = getattr(obj, im_func)
     except AttributeError:  # Python 2.6
```
mmckerns, line 1874: this is marked as python 2.6, so it might be able to go away.  Not sure.

dill/_dill.py
```diff
@@ -2154,7 +1902,7 @@ def save_function(pickler, obj):
                 found, _ = _getattribute(module, obj.__qualname__)
                 if getattr(found, '__func__', None) is obj:
                     _pypy_builtin = True
-            except:
+            except AttributeError:
```
mmckerns, line 1905: Not sure if AttributeError is only possibility here...

dill/__init__.py
```diff
@@ -42,27 +42,28 @@
 
 ``dill`` is part of ``pathos``, a python framework for heterogeneous computing.
 ``dill`` is in active development, so any user feedback, bug reports, comments,
-or suggestions are highly appreciated.  A list of issues is located at https://github.com/uqfoundation/dill/issues, with a legacy list maintained at https://uqfoundation.github.io/project/pathos/query.
```
mmckerns, line 45: Ok, this is mildly annoying, and hopefully temporary, but if you modify the module `__doc__` in this file, you also need to modify it on the README.md.  Hopefully, I'll reduce this to maintaining a single copy of the doc soon.  The reason for the two copies that I needed the doc in the different formats... and it wasn't as easy as it is now to convert between the formats.

mmckerns (changes requested):
Generally LGTM. Thanks for all the hard (and tedious) work.  Please address all of my comments and change requests. One thing to also watch out for in `dill._dill` is cases where we want to always want to trap any exception, regardless if it looks like there's only a particular type that is possible (i.e. the exception is used more like a if/else).  I noted a few cases that I couldn't remember fully what scope the exception block was intended to have.

dill/_dill.py
mmckerns: If you look at the part you converted to (...), I believe that `vars` acts more like `locals()` than `globals()` (so it can be masked).

leogama: Of course, `vars()` without arguments is equivalent to `locals()`, and the restrictions that apply to `locals()` also apply **in this case.** But for `vars(obj)` the behavior is strictly equivalent to a function defined as:

```python
def vars(obj):
    if not hasattr(obj, '__dict__'):
        raise TypeError
    return obj.__dict__
```

The implementation in CPython is just like it: https://github.com/python/cpython/blob/bbcf42449e13c0b62f145cd49d12674ef3d5bf64/Python/bltinmodule.c#L2404-L2423

dill/__init__.py
leogama: OK, applied the diff to it

dill/_dill.py
leogama: So I'll update everything to `vars(module)`, OK?

dill/_dill.py
leogama: You are right, I missed that there was an implicit else statement. I'll will just convert the last `elif` to an `else` then.

dill/_dill.py
leogama: Maybe put it next to the `and ("w" in mode or "x" in mode):` line below?

dill/_dill.py
leogama: I thought it would raise an `AttributeError` if there wasn't "cell contents", but it actually rose `ValueError` when the cell was empty. I've just found the documentation about this and it seems to be the case (see with `help(CellType)`):

```
Help on class cell in module builtins:

class cell(object)
 |  Create a new cell object.
 |  
 |   contents
 |     the contents of the cell. If not specified, the cell will be empty,
 |     and 
 |  further attempts to access its cell_contents attribute will
 |     raise a ValueError.
(...)
```

dill/_dill.py
leogama: Missed this one.

dill/_dill.py
leogama: From [this old SO post](https://stackoverflow.com/a/1677671/3738764), it seems that `classmethod`s didn't have a `__func__` attribute before Py2.7. I think it's safe to remove.

mmckerns: Actually I remember that now. :)

dill/_dill.py
mmckerns: I wouldn't unless you are getting a speed bonus out of it -- otherwise the change just seems to be semantic sugar.  (I'm also a bit hesitant about some of the error specification you did as you might be able to tell from my comments.)

leogama: > I wouldn't unless you are getting a speed bonus out of it -- otherwise the change just seems to be semantic sugar.

As you wish. And oops! `vars(module)` is actually slower... OK, so I won't change anything, but I'll try to keep consistency with the submodule style in future PRs.

> (I'm also a bit hesitant about some of the error specification you did as you might be able to tell from my comments.)
Yes, I'm reviewing these carefully. I really wasn't sure about some of the specific errors, but I left them on purpose for you to review and eventually clarify if they were appropriate. If we can't agree on whether a exception list cover all the expected cases (not including misuses that should rise an error), I'll change it to `except Exception`, as I did with other `try` blocks for which there was no clue of the possible error types.

> There's a lot of changes in this PR, so we need to make sure this gets well tested.
Yeah... I may have diverted a bit from the main goals. Let's wrap it up.

mmckerns: > I'll change it to except Exception, as I did with other try blocks for which there was no clue of the possible error types.

This is better... as it's a change in how something pickles if the error is no longer caught, but maybe gets caught elsewhere.

dill/_dill.py
leogama: Let's see:
- `obj.__qualname__` may rise an `AttributeError` (AE);
- `_getattribute()` only explicitly raises AE;
- `module` may be a module or `None` (from `_import_module(safe=True)`), and in this case it will generate an AE inside `_getattribute()`;
- if `obj.__qualname__` is not a string (which would by wild, as you cannot set `__qualname__` to anything but a string), it will rise an AE if it doesn't have a `split()` method;
- only if `obj.__qualname__` is of another type that also has a `split()` method, like a `bytes` object, it will rise a `TypeError` if the method expects an argument of a different type (also `bytes` for a `bytes` object).

I would argue that in this last case raising an exception is appropriate.

leogama:
Applied the requested corrections and then merged with master HEAD. All tests passing 100%. I'm feeling this is ready to go!

leogama:
> Generally LGTM. Thanks for all the hard (and tedious) work.

You're welcome! 🙂 Applying one step at a time to all files sped up the process. Vim search and substitutions also helped a lot. But I'm pretty sure that reviewing for you was as much as tedious as editing for me...

> One thing to also watch out for in `dill._dill` is cases where we want to always want to trap any exception, regardless if it looks like there's only a particular type that is possible (i.e. the exception is used more like a if/else). I noted a few cases that I couldn't remember fully what scope the exception block was intended to have.

In the end, except for the `ImportError`s, I left most cases as catch-all `except Exception` clauses. Either because I spotted they were a general error trap or because I had no idea of the kind of error expected. However, even if it stay as it is, it would be great if you could annotate the ones you still remember the scope, no matter how vague. Sometimes it is possible to infer what kind of things can go wrong based on the context, but some generic exception blocks seem like real mysteries.

mmckerns:
> it would be great if you could annotate the ones you still remember the scope, no matter how vague.

Some of this code is 15 years old. I feel like I should get into a time machine and tell younger me to document the exception blocks more.

mmckerns:
> Applying one step at a time to all files sped up the process.

Side note: can you email me a list of the keywords you searched for in your removal of dead code?  I'm making similar changes on the other UQF codes.

mmckerns:
Thanks, that's similar to my list.  I've also included `attr(` and `version` and `0x2'/'0x3'.  I often used `getattr` for 2/3 compatibility.

mmckerns (approved):

leogama:
@mmckerns: just pushed the merge commit with the trace PR.

Issue #485, opened 2022-05-24
Discussion: cPickle


leogama:
Hey, I've been working on a prototype of a "portable pickle" feature and started wondering whether it could eventually work with cPickle (or `_pickle`), the accelerator module written in C. I've also seen some comments in the code referencing the possibility that dill in general could use it.

However, after doing some investigation I noted that, even if the `save_<type>` functions are updated to use state setters (Py 3.8+) and don't rely on internal Pickler methods like `save` and `save_reduce`, there remains a big problem: cPickle will always save types and functions as globals (there isn't an internal `dispatch` table that we can modify as with the Python implementation).

I thought of creating a C extension module to play with cPickle internals, but all the C functions are private. The only remaining alternative I see is to have a complete clone of `Pickler` code with some minor tweaks to allow overriding these types' saving.

Have you also wondered about these issues? Do you think it is feasible to have a C extension compatible with multiple Python 3 versions? (I have no ideia how building and distributing these works.)

anivegesana:
It is doable. You ship a precompiled binary for every single OS/Python version pair that you want to support. There aren't way too many combinations, just 20, plus the one for noarch for a fallback. The reason that I didn't bother thinking about it is because it would require some cleverness to build the C extension around the Python package in a way that would reduce code duplication but result in fast performance.

This causes an issue for people who want to use pickle5 in older versions of Python. The pickle5 package has a C library clone of pickle as well like you described, but it is not as trivial to get it working with a hypothetical C-dill compared to the current Python implementation.

If you would like to pursue this, I would say make a clone of cPickle that has all of its functions exposed and we can extend from it if it if the C extension is present.

mmckerns:
> Do you have stats on how many IronPython or Jython users dill has? I do not even know if there is support for them or if they are still maintained.

As far as I know, they are both supported.  The most detailed stats I track are here: https://pypistats.org/packages/dill, which doesn't give details on Jython/IronPython.

anivegesana:
I just noticed that they have no stable Python 3 builds, so they are probably implicitly unsupported now that Python 2 is no longer supported. The only binaries that need to be built are for CPython, which makes this problem much easier.

leogama:
[...]

Issue #19, opened 2013-12-27
function pickles are only quasi-deterministic, does this matter?


stepleton-xx:
Pickles of functions (and pickles of things that contain functions, like classes) are not quite deterministic---they depend on iteration order of the `_reverse_typemap` inside dill.py. Depending on the order, either the symbol "LambdaType" or "FunctionType" will be used to represent functions. Either will work as far as unpickling goes, but having different representations of the same value can cause trouble with e.g. caching.

While most invocations of Python 2.x yield the same iteration order for `_reverse_typemap`, use of the -R flag (recommended for user-facing services; c.f. http://www.ocert.org/advisories/ocert-2011-003.html) randomizes this order.

Note that the functionality of -R is on by default for versions >= 3.3:
http://docs.python.org/3/whatsnew/3.3.html

davips:
Any news regarding providing determinism for dill? I am using it only for hashing, so  are there any better workaround than sorting the dump by byte?

anivegesana:
Are you using Python 3.7+? Because dictionaries are now insertion order, for any particular version of Python after 3.6, the type dictionary should be deterministic, and this issue no longer applies. Since dill 0.3.6 will no longer support Python 2.7, this issue should no longer be relevant if I am not mistaken.

davips:
Is is still nondeterministic in python 3.8.10:
```python
In [3]: import dill as d; from hashlib import md5; f = lambda x: x + 1; dum = d.dumps(f); md5(dum).hexdigest()
Out[3]: 'bbe58ab30aec9b1feb80cd3e4b2b7024'

In [4]: import dill as d; from hashlib import md5; f = lambda x: x + 1; dum = d.dumps(f); md5(dum).hexdigest()
Out[4]: '3fc384d6edf117061f781113a60fd053'
```

mmckerns:
I'm not seeing the same behavior in 3.8.13.

```
Python 3.8.13 (default, May 10 2022, 11:26:38) 
[Clang 10.0.1 (clang-1001.0.46.4)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import dill
>>> dill.settings['recurse'] = True
>>> from hashlib import md5
>>> f = lambda x:x+1
>>> dum = dill.dumps(f)
>>> md5(dum).hexdigest()
'65e4cf11903c322b610b68ddb7c71283'
>>> f = lambda x:x+1
>>> dum = dill.dumps(f)
[...]

PR #501, opened 2022-06-07
Deterministic mode

anivegesana:
Implements https://github.com/uqfoundation/dill/issues/19#issuecomment-1141528750

For now, this is what I have chosen for the behavior of "deterministic" mode:
- `set` and `frozenset` will be sorted before being pickled.
  - Subclasses of `set` and `frozenset` will not be effected (and will remain nondeterministic) because they can implement their own `__reduce__` functions which don't have to follow the conventions of `set`'s pickling procedure.
  - If the elements are incomparable (e.g. `complex`), they will be sorted by their hash instead. This will not create a natural order of elements that is easy to understand, but if the `__hash__` function of the class doesn't depend on `id`, it will be deterministic.
  - If using the faster cPickle based pickler outlined in #485, this feature may be disabled.
- `dict` and subclasses will remain pickled in insertion order.
  - Entries in global variable dictionaries will be in order for each function. The dictionary as a whole, however, will be ordered in visitation order by function and will not be sorted in alphabetical order. This will mean that the globals dictionaries will be deterministic given that the visitation order of functions is deterministic.
  - This feature is guaranteed.
- `code` objects will **not** have their line numbers removed and the file name and function name will not be modified.
  - I thought this would be best to leave up to the packages that use dill rather than requiring all packages that want determinism to drop line number information from code.
  - For now, huggingface datasets can override `save_code` like they already do, but issues like #488 would make that difficult to maintain and it would be best to try to find a way to emit information that would help users selectively zero out/ignore line information in code objects when hashing. That way, information that debuggers need still remains in the pickle, but the hash remains consistent.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,6 +117,16 @@ as keyword arguments to several ``dill`` functions:
   remote system which does not have the original file on disk. Options are
   *HANDLE_FMODE* for just the handle, *CONTENTS_FMODE* for the file content
   and *FILE_FMODE* for content and handle.
+* with *deterministic=True*, dill will try to make pickles more likely to
+  be the same if an object is pickled multiple times. Currently, here is
+  the feature set:
+  * `set` and `frozenset` will be sorted before being pickled.
+    * Subclasses of `set` and `frozenset` will not be effected (and will remain nondeterministic) because they can implement their own `__reduce__` functions which don't have to follow the conventions of `set`'s pickling procedure.
+    * If the elements are incomparable (e.g. `complex`), they will be sorted by their hash instead. This will not create a natural order of elements that is easy to understand, but if the `__hash__` function of the class doesn't depend on `id`, it will be deterministic.
+    * If using the faster cPickle based pickler outlined in [#485](https://github.com/uqfoundation/dill/issues/485), this feature may be disabled.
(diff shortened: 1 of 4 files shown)


anivegesana:
I am having a difficult time testing this because non-determinism is an implementation specific detail, so CPython and PyPy have different data structures for sets and it is hard to find a test case that doesn't work on both of them.

anivegesana:
Hey @AaronFriel,

I haven't looked at this in forever and will need some time to figure out what this did and tried to accomplish. With a couple of pass-throughs, I think most of the functionality can be obtained by registering custom pickling functions for set. Is there a particular use case you are looking to use this for?

AaronFriel:
Pulumi, as an infrastructure as code tool, computes diffs between cloud provider resources declared in code, such as an AWS Lambda function, to determine whether to update the resource. A deterministic pickle ensures that we don't see spurious updates.

N-Demir:
+1 would really love to see a deterministic pickle because so many lib's caching strategies depend on it (like prefect)

JGrel:
How is it going with deterministic mode?

faaany:
Any update to support deterministic mode?

PR #507, opened 2022-06-10
Fix load_session() and restrict loading a session in a different module

### LINKED ISSUE #498: Reloading a session into the same module keeps the previous session
I'm generating 2 session dumps with

```
import dill
a = 1234
dill.dump_session('test1.session')
```

and

```
import dill
b = 1234
dill.dump_session('test2.session')
```

Then loading them back into a module sequentially

```
import types
import dill

mod = types.ModuleType('test')
print(mod.__dict__.get('a', None))
print(mod.__dict__.get('b', None))

dill.load_session('test1.session', main=mod)
print(mod.__dict__.get('a', None))
print(mod.__dict__.get('b', None))


mod = types.ModuleType('test')
print(mod.__dict__.get('a', None))
print(mod.__dict__.get('b', None))

dill.load_session('test2.session', main=mod)
print(mod.__dict__.get('a', None))
print(mod.__dict__.get('b', None))
```

I get the following output

```
None   # a
None   # b
1234   # a
None   # b
None   # a
None   # b
1234   # a
1234   # b
```

Before the second session file is read, it clearly shows that the mod variable has no a nor b set. But then when reading in the test2.session file it suddenly sets both a and b. But only b was actually dumped into that file.

Is this an expected behavior?
I'd have expected that after the test2.session was loaded, only the b variable to be present

leogama:
Resolves #498. Ongoing work, I'm adding a fix/feature per commit. The plan:

Operations allowed by `load_session()`:
- Save an importable module and restore to the same module
- Save a module created at runtime (with `ModuleType()`) and restore to another runtime module with the same name
- ~Save an importable module and restore to a runtime created module with the same name (allow unpickling multiple sessions, for comparing versions of objects for example).~

Operations **disallowed** by `load_session()`:
- Save any module and restore it to a module with a different name

Other changes:
- The `main` parameter is now also optional for simply loading modules other than `__main__`
- The `main` parameter accepts a runtime module, or a name for `load_session()` to create it with `ModuleType(main)`
- If the `main` module is other than `__main__`, it is returned:
```python
module = dill.load_session('module_session.pkl')  # don't need to "import module"
```


leogama:
@mmckerns, I'm progressing with this issue, but I can't find a way to allow a user to restore a common session into a module created with `ModuleType()` without also restoring it to the original module or without breaking the API...

mmckerns:
> I can't find a way to allow a user to restore a common session into a module created with ModuleType() without also restoring it to the original module or without breaking the API.

That's ok. It should be a pretty small corner case.  Some options are: (1) don't allow it, (2) allow it but also restore the original module where (a) the original module is imported then removed, (b) the original module is imported with an alias... or (c) something else.  I'd look at the `importlib` for some guidance on the behavior people might expect, and how that behavior is implemented.

leogama:
> > I can't find a way to allow a user to restore a common session into a module created with ModuleType() without also restoring it to the original module or without breaking the API.
> 
> That's ok. It should be a pretty small corner case. Some options are: (1) don't allow it, (2) allow it but also restore the original module where (a) the original module is imported then removed, (b) the original module is imported with an alias... or (c) something else. I'd look at the `importlib` for some guidance on the behavior people might expect, and how that behavior is implemented.

What I ended doing is to create a separate function `load_session_copy()` that loads the original module, makes a backup copy of its state, calls `load_session()`, copies the loaded dictionary to a new module and restores the original module.

The problem with the options 2.a and 2.b that you suggested is that the CPython 3.7 import system seems to be somewhat broken. As I mentioned in this comment (https://github.com/uqfoundation/dill/pull/463#issuecomment-1112677576) and a couple ones before it, the sequence of commands `import module` → `del sys.modules['module']` → `import module` doesn't always put the module back into `sys.modules` (and there's also the problem of submodules imported indirectly).

The `importilb` docs didn't give me much light, but I used it as reference to clear out the attributes `__loader__`, `__spec__` and `__path__` in the created runtime module.

leogama:
I've found at least one package that would benefit from this PR: https://github.com/SAME-Project/same-project/blob/main/sameproject/ops/execution.py

dill/_dill.py
```diff
@@ -504,29 +505,138 @@ def dump_session(filename='/tmp/session.pkl', main=None, byref=False, **kwds):
...
+        try:
+            del sys.modules[runtime_main]
+        except (KeyError, NameError):
+            pass
+    assert module is main
+    _restore_modules(unpickler, module)
+    if module is not _main_module:
+        return module
+
+def load_session_copy(filename='/tmp/session.pkl', **kwds):
```
mmckerns, line 606: I don't like the name of this function.

dill/_dill.py
```diff
@@ -504,29 +505,138 @@ def dump_session(filename='/tmp/session.pkl', main=None, byref=False, **kwds):
...
+        file = open(filename, 'rb')
+    if not hasattr(file, 'peek'):
+        try:
+            import io
+            file = io.BufferedReader(file)
+        except Exception:
+            pass  # ...and hope for the best
+    return file
+
+def _inspect_pickle(file, main_is_none):
```
mmckerns, line 525: This function seems badly named. Can you adjust to better repent the result of the function?

dill/_dill.py
```diff
@@ -504,29 +505,138 @@ def dump_session(filename='/tmp/session.pkl', main=None, byref=False, **kwds):
...
-        if f is not filename:  # If newly opened file
-            f.close()
+        if file is not filename:  # if newly opened file
+            file.close()
     return
 
-def load_session(filename='/tmp/session.pkl', main=None, **kwds):
-    """update the __main__ module with the state from the session file"""
-    if main is None: main = _main_module
+def _open_peekable(filename):
```
mmckerns, line 512: The point of this function is to try to ensure the file is in a mode where one can `peek` without advancing?

mmckerns (changes requested):
Please address the review comments.

dill/_dill.py
mmckerns: Maybe a better choice is to rename it `_load_session`, then have it called seamlessly from `load_session` if someone sets a `kwarg` (where the default `None` would be `__main__`).

leogama:
I did various changes in the commit "adjustments":

- Wrote documentation for the public functions and added some Sphinx options to `docs/source/conf.py` that shouldn't affect already existing documentation (at least I didn't notice any changes using the RTD theme).
- Renamed private function names and `load_session_copy()` to `load_vars()`, which now returns only the module's dictionary, as it shouldn't be used as a restored module because of the problems we discussed earlier. See if you like it this way or if you still prefer it merged with `load_session()`. I added an option to use the current state of the module being restored as a base namespace, before loading objects from the file.
- Converted the old `_open_pickable` function into a wrapper class `_PeekableReader`, which I think makes the code purpose more transparent.

dill/_dill.py
```diff
@@ -483,8 +488,53 @@ def _restore_modules(unpickler, main_module):
         pass
 
 #NOTE: 06/03/15 renamed main_module to main
-def dump_session(filename='/tmp/session.pkl', main=None, byref=False, **kwds):
-    """pickle the current state of __main__ to a file"""
+def dump_session(
```
mmckerns, line 406: This is nice. Two potentially controversial suggestions/thoughts:

1. rename `load_session` to something like `load_module`. Is there a better name for it now, given the current functionality?
2. rename `byref`, or integrate this `byref` with `settings['byref']`. It's confusing as is.

dill/_dill.py
```diff
@@ -509,20 +559,42 @@ def dump_session(filename='/tmp/session.pkl', main=None, byref=False, **kwds):
...
+        try:
+            return io.BufferedReader(stream)
         except Exception:
-            pass  # ...and hope for the best
-    return file
+            pass
+    return _PeekableReader(stream)
 
-def _inspect_pickle(file, main_is_none):
+def _identify_session_module(file, main_is_none):
```
mmckerns, line 513: Ok, this looks good. I'm going to ask you to think about naming again. How about `_identify_main` or `_identify_main_module` or `_identify_module`?  Also, the arg name `main_is_none` is not good... what is the intent of this arg?

dill/_dill.py
```diff
@@ -534,25 +606,81 @@ def _inspect_pickle(file, main_is_none):
...
+        # ValueError occours when the end of the chunk is reached (without a STOP).
+        if isinstance(error, NotImplementedError) and not main_is_none:
+            # file is not peekable, but we have main.
             return None
         raise UnpicklingError("unable to identify main module") from error
 
-def load_session(filename='/tmp/session.pkl', main=None, **kwds):
-    """update the __main__ module with the state from the session file"""
-    file = _open_peekable(filename)
+def load_session(
```
mmckerns, line 536: Same comment here on naming as `dump_session`.

dill/_dill.py
```diff
@@ -603,40 +731,74 @@ def load_session(filename='/tmp/session.pkl', main=None, **kwds):
     if module is not _main_module:
         return module
 
-def load_session_copy(filename='/tmp/session.pkl', **kwds):
+def load_vars(
```
mmckerns, line 651: The name is much better.  Also consider: `load_namespace` and `load_module_vars`.  Fundamentally, the difference here between `load_session` and `load_vars` is the return object... right? The former returns as a module (or main), and the latter as a dict.  So... they should be either very closely named, or this function should begin with an underscore and be available from `load_session` through the use of a keyword argument.

mmckerns (commented):
The code really LGTM.  However, before we merge... let's iterate once or twice more purely on the names of the functions I've identified.

dill/_dill.py
leogama: Oh, great. I was trying to close the review I had opened in the beginning and ended up deleting all my last comments...

1. Agree. It's the best name and a way to change the API without breaking it.
2. Yes, but we need to choose one more name. Some ideas:

- `refer_imported`
- `disown_imported`
- `link_imported`
- `imported_byref`

dill/_dill.py
leogama: Rewriting my comment: `load_module_vars()` is preferable as it's parameters may diverge from the ones from `load_module()` as the time goes. There's already a new `update` parameter for it. Plus its functionality is discoverable when tab-completing `dill.load_m...` (without disrupting the completion).

I like `load_namespace`, but it can trick users into thinking it returns a `types.SimpleNamespace` object or something else.

dill/_dill.py
leogama: Rewriting comment: I prefer `_identify_module()`. And `main_is_none` was a poor choice. Changed it.

leogama:
I changed the `modmap` object to use `types.SimpleNamespace` instead of that `AttrDict` class. But I still think that a custom dict subclass for `settings['module']` would be very handful. It would allow:

- access of items via attributes
- lookup default value in parent dictionary (if set)
- automatic creation of empty parent for submodules settings

The settings API I'm aiming at is somewhat as this:
```python
>>> from dill import FilterRules, settings
>>> mod_settings = settings['module']
>>> mod_settings['imported_byref']  # item access
False
>>> mod_settings.imported_byref = True  # attribute access
>>> mod_settings.__main__ = {'imported_byref': False}  # a module specific session setting
>>> mod_settings.__main__.imported_byref  # instead of mod_settings['__main__']['imported_byref']
False
>>> mod_settings['my.submodule'] = {'rules': FilterRules()}  # create a 'submodule' entry under a 'my' entry
>>> mod_settings.my.submodule.rules.exclude.add('somevar')
```

dill/_dill.py
mmckerns: I like `dump_module` and `load_module`, but don't like `imported_byref`.  Any one word suggestions to replace `byref` here? How about `refimport`? `refmod`? `refmodule`?

dill/_dill.py
mmckerns: > it can trick users into thinking it returns a types.SimpleNamespace object or something else.

Ah... you are right.  Hmm. `load_module_asdict`?  `load_asdict`? `load_dict`?   Hmm.... I actually think I like `load_vars` better than `load_module_vars`.  I'm not sure about what this should be called yet.

dill/_dill.py
mmckerns: I do think the repeated use of `module` for many of the related functions is really nice.

mmckerns:
I see... so you want to (for example) have a settings "tree" as opposed to a flat settings dict.  This might take a bit of design discussion. How about we tackle that in a separate PR?

leogama:
> I see... so you want to (for example) have a settings "tree" as opposed to a flat settings dict. This might take a bit of design discussion. How about we tackle that in a separate PR?

I'm refining the idea. Indeed, a settings "tree" as I was imagining may have an implementation too complex. I'm thinking now of having just the filter rules for `dump_module()` as a tree, by subclassing `FilterRules` (or `ExcludeRules`) and some mapping type.

However, I'm still thinking of a format for a config file, including the filter rules for specific modules. Maybe we should use `configparser` (the format and the reader) as a base:

```ini
[dill]
byref = true
protocol = 3

[dump_module]
exclude.types = [FunctionType, ModuleType, TypeType]  # or [function, module, type]

[dump_module.__main__]
exclude.regexes = "_.*"
```

dill/_dill.py
leogama: What about `refimported`? Because the option means:

> _Pickle all the imported objects by reference._

Even `refimport` makes me think of modules, and the option is not specific to modules.

By the way, I chose `imported_byref` semi-randomly to substitute `byref` in the last commit.

dill/_dill.py
leogama: My favorites, in no specific order, are `load_module_asdict`, `load_module_vars` and `load_vars`. The other seem either nonspecific or possibly misleading. Let's leave it hanging for a while.

A short description of what the function does is:

> "The equivalent of
> ```python
> lambda file: vars(load_module(file)).copy()
> ```
> but without touching the original module."

---
If you prefer... 😜 

> ```python
> lambda file: load_module(file).__dict__.copy()
> ```

mmckerns: `load_module_asdict`: we have a winner then.  ...and the first short description is perfect.

dill/_dill.py
mmckerns: `refimported` sound good to me

leogama: As you said the other day, I don't love it, but is Good Enough™

leogama:
Oops! My two last messages were intended to go to #475...
I finished renaming, documenting and fixing a couple things here.

mmckerns (approved):
The documentation could use going over once more to clean it up a bit, but otherwise **LGTM**.  If you feel like you don't want to make another pass at the documentation, I'll pull it and then tweak the docs a bit myself.

leogama:
> The documentation could use going over once more to clean it up a bit

I'm sorry? Didn't understand a word. I'm happy by now. Merge it, please, so you can tweak it as you like, and I can merge it with the filter PR and finish that. I'm going to complement the code annotation and the private functions' documentation there if necessary.

Or... Can't you add suggested changes directly here, in a way that I could just accept and include in a new commit. I think I saw a button in the desktop web interface, in the comment box.

PR #508, opened 2022-06-10
Update method related types to CPython 3.7 and PyPy 3.7

leogama:


leogama:
@mmckerns, it seems that PyPy changed its types and I need these to make #475 pass on Travis.

PR #513, opened 2022-06-18
A hook that allows dill to save the versions of Python and dill

anivegesana:
This may be useful later in preventing issues like #488 from occurring in the future. By saving the version of Python and dill in the pickle, we can save some trouble later. I am also OK with removing the dill version, since, with careful planning, we can make sure that we never need to use it. This will add at least 50 bytes to very pickle, so maybe an option to turn it off would be desirable.

@mmckerns thoughts?

### DIFF
--- a/dill/__init__.py
+++ b/dill/__init__.py
@@ -290,6 +290,9 @@ from ._dill import dump, dumps, load, loads, dump_session, load_session, \
     PicklingWarning, UnpicklingWarning
 from . import source, temp, detect
 
+_dill._dill = _dill
+_dill.__version__ = __version__
+
 # get global settings
 from .settings import settings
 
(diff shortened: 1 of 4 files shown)


mmckerns:
My thought is that it would simplify much of the logic when needing to identify the version of python used to serialize an object.  I believe that we do that really only when python has made a backward-incompatible change -- which isn't too often (except for in `CodeType` objects). So, I'm not convinced that it's worth the increase to the size of every pickle (primarily why I didn't save the python version in `args` in `_save_code`).

mmckerns:
Nice idea though.

anivegesana:
I will close this for now. We can resurrect this later if we find a need to store stateful information inside the unpickler.

PR #464, opened 2022-04-23
Fix dill.source.getsource compatibility for IPython interactive session.

VincentDary:
I propose a little fix for `dill.source.getsource` for IPython compatibility. It allows to retrieve function and class source code in IPython interactive session. The fix solves the following issue in IPython shell.

```
In [1]: import dill

In [2]: def foo():
   ...:     return 0
   ...: 

In [3]: class Bar:
   ...:     pass
   ...: 

In [5]: dill.source.getsource(foo)
---------------------------------------------------------------------------
OSError                                   Traceback (most recent call last)
Input In [5], in <cell line: 1>()
----> 1 dill.source.getsource(foo)

File ~/.local/lib/python3.10/site-packages/dill/source.py:354, in getsource(object, alias, lstrip, enclosing, force, builtin)
    352 # get source lines; if fail, try to 'force' an import
    353 try: # fails for builtins, and other assorted object types
--> 354     lines, lnum = getsourcelines(object, enclosing=enclosing)
    355 except (TypeError, IOError): # failed to get source, resort to import hooks
    356     if not force: # don't try to get types that findsource can't get

File ~/.local/lib/python3.10/site-packages/dill/source.py:325, in getsourcelines(object, lstrip, enclosing)
    312 def getsourcelines(object, lstrip=False, enclosing=False):
    313     """Return a list of source lines and starting line number for an object.
    314     Interactively-defined objects refer to lines in the interpreter's history.
    315 
   (...)
    323     If lstrip=True, ensure there is no indentation in the first line of code.
    324     If enclosing=True, then also return any enclosing code."""
--> 325     code, n = getblocks(object, lstrip=lstrip, enclosing=enclosing, locate=True)
    326     return code[-1], n[-1]

File ~/.local/lib/python3.10/site-packages/dill/source.py:251, in getblocks(object, lstrip, enclosing, locate)
    241 def getblocks(object, lstrip=False, enclosing=False, locate=False):
    242     """Return a list of source lines and starting line number for an object.
    243     Interactively-defined objects refer to lines in the interpreter's history.
    244 
   (...)
    249     DEPRECATED: use 'getsourcelines' instead
    250     """
--> 251     lines, lnum = findsource(object)
    253     if ismodule(object):
    254         if lstrip: lines = _outdent(lines)

File ~/.local/lib/python3.10/site-packages/dill/source.py:154, in findsource(object)
    151         lines = linecache.getlines(file)
    153 if not lines:
--> 154     raise IOError('could not extract source code')
[...]

PR #516, opened 2022-06-20
Fix dill.source.getsource compatibility for IPython interactive session.

VincentDary:
Come back of the https://github.com/uqfoundation/dill/pull/464, sorry I didn't think you would be interested.

### DIFF
--- a/dill/source.py
+++ b/dill/source.py
@@ -27,6 +27,7 @@ import re
 from inspect import (getblock, getfile, getmodule, getsourcefile, indentsize,
                      isbuiltin, isclass, iscode, isframe, isfunction, ismethod,
                      ismodule, istraceback)
+from inspect import getsource as inspect_getsource
 from tokenize import TokenError
 
 from ._dill import PY3
@@ -127,8 +128,14 @@ def findsource(object):
                 err += ", please install 'pyreadline'"
         if err:
             raise IOError(err)
-        lbuf = readline.get_current_history_length()
-        lines = [readline.get_history_item(i)+'\n' for i in range(1,lbuf)]
+
+        if hasattr(module, 'get_ipython'):
+            if isfunction(object):
+                return [''.join(inspect_getsource(object))], 0
+            lines = module.get_ipython().history_manager.input_hist_raw
+        else:
+            lbuf = readline.get_current_history_length()
+            lines = [readline.get_history_item(i)+'\n' for i in range(1,lbuf)]
     else:
         try: # special handling for class instances
             if not isclass(object) and isclass(type(object)): # __class__


mmckerns:
I'm very interested.

VincentDary:
Ok, I'm no familiar with `jupyter` but I will check this compatibility issue.

mmckerns:
>  Standard Library inspect.getsource() actually works on IPython and Jupyter

Interesting. I just tested this... and can confirm it does
```
Python 3.7.13 (default, May 10 2022, 11:13:40) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.30.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import inspect

In [2]: def foo(x):
   ...:     return x*foo(x)
   ...: 

In [3]: inspect.getsource(foo)
Out[3]: 'def foo(x):\n    return x*foo(x)\n'

```

That's good, as it didn't used to work. You'd get results like you get from a python shell:
```
>>> import inspect
>>> 
>>> def foo(x):
...   return x*x*x
... 
>>> inspect.getsource(foo)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/inspect.py", line 973, in getsource
    lines, lnum = getsourcelines(object)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/inspect.py", line 955, in getsourcelines
    lines, lnum = findsource(object)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/inspect.py", line 786, in findsource
    raise OSError('could not get source code')
OSError: could not get source code

```
> Why is it better to get the source file from the object's module instead of the object directly like inspect.findsource() does? Looking at the inspect.getfile() code it appears that the heuristics used there won't work for all types of objects, but shouldn't getting the file through the module be just a fallback method? If yes, then changing this could make inspection work out of the box for IPython and Jupyter, with no need to mess around with the history there.

It's my understanding that `getsource` uses `getsourcelines` which uses `findsource`, where `findsource` retrieves the module's code and the linenumber of where the code starts.  Some of the code from `dill.souce` is very similar to what is in `inspect`.

So, we should trace down why this happens, and see to what depth in the code we can leverage what `inspect` is doing without losing functionality in `dill.source`.

leogama:
> It's my understanding that `getsource` uses `getsourcelines` which uses `findsource`, where `findsource` retrieves the module's code and the linenumber of where the code starts. Some of the code from `dill.source` is very similar to what is in `inspect`.

That's exactly what I've found. However, `inspect` gets the source file directly from the object, while `dill.source` first tries to get the object's module and then the module's file. I guess that generally the module's file and the object's file (retrieved from `func.__code__.co_filename` in the case of functions) are the same, but in this special case they are not:

```python
In [1]: import inspect, __main__
In [2]: def f(): pass
In [3]: f.__module__ 
Out[3]: '__main__'
In [4]: inspect.getfile(f)
Out[4]: '<ipython-input-2-40d3b0035d30>'
In [5]: inspect.getfile(__main__)
---------------------------------------------------------------------------
(...)
TypeError: <module '__main__'> is a built-in module
```

The question is: why does `dill.source` diverges from `inspect` and tries to retrieve the source file from the object's module? Is it because it covers more types of objects? And, more important, is it safe to use this strategy just as a fallback, or would it break some other corner case?

mmckerns:
I'm closing this, as there's a duplicate PR (#531) that also appears to work in `jupyter`.

Issue #518, opened 2022-06-25
fix object classification in _objects


mmckerns:
`dill._objects` categorizes objects into three dicts:
```
# objects used by dill for type declaration
registered = d = odict()
# objects dill fails to pickle
failures = x = odict()
# all other type objects
succeeds = a = odict()
```

Apparently, this has not been updated for recent changes:
```
>>> import dill
>>> from dill._objects import failures, registered, succeeds
>>> import warnings
>>> warnings.filterwarnings('ignore')
>>> def check(d, ok=True):
...   res = []
...   for k,v in d.items():
...     try:
...       z = dill.copy(v)
...       if ok: res.append(k)
...     except:
...       if not ok: res.append(k)
...   return res
... 
>>> check(failures) # should be empty
['SetIteratorType', 'DictionaryItemIteratorType', 'DictionaryKeyIteratorType', 'DictionaryValueIteratorType', 'MethodCallerType', 'DictItemsType', 'DictKeysType', 'DictValuesType', 'RawTextHelpFormatterType', 'RawDescriptionHelpFormatterType', 'ArgDefaultsHelpFormatterType', 'PyCapsuleType']
>>> check(registered, ok=False) # should be empty
[]
>>> check(succeeds, ok=False) # should be empty
['CLibraryLoaderType', 'ConnectionType', 'CursorType', 'BZ2FileType', 'BZ2CompressorType', 'BZ2DecompressorType', 'DialectType', 'SocketType', 'SocketPairType', 'MemoryIteratorType', 'SymtableEntryType', 'GzipFileType']
>>> import builtins
>>> import types
>>> q = dill._dill._reverse_typemap
>>> p = {k:v for k,v in q.items() if k not in vars(builtins) and k not in vars(types)}
>>> set(p.keys()).difference(registered.keys()) # should be empty
{'OdictItemsType', 'DictKeysType', 'PyBufferedWriterType', 'PyCapsuleType', 'PyBufferedRandomType', 'OdictValuesType', 'DictValuesType', 'BufferedRandomType', 'SymtableStentryType', 'TextWrapperType', 'PyTextWrapperType', 'BufferedWriterType', 'DictItemsType', 'OdictKeysType', 'PyBufferedReaderType', 'FileType', 'BufferedReaderType'}
>>> set(registered.keys()).difference(p.keys())
{'ClassMethodDescriptorType', 'MethodWrapperType', 'MethodDescriptorType', 'WrapperDescriptorType'}
```

PR #519, opened 2022-06-25
correct failures,  registered, and succeeds

mmckerns:
1) move failing types from succeeds to failures, and succeeding types from failures to succeeds
2) update registered so matches reverse_typemap not in (builtins and types)

fixes: #518

### DIFF
--- a/dill/_dill.py
+++ b/dill/_dill.py
@@ -747,11 +747,13 @@ if InputType:
     _incedental_reverse_typemap['InputType'] = InputType
     _incedental_reverse_typemap['OutputType'] = OutputType
 
+'''
 try:
     import symtable
     _incedental_reverse_typemap["SymtableEntryType"] = type(symtable.symtable("", "string", "exec")._table)
-except:
+except: #FIXME: fails to pickle
     pass
+'''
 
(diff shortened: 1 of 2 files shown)


mmckerns:
After the above commits...
```
>>> import dill
>>> from dill._objects import failures, registered, succeeds
>>> import warnings
>>> warnings.filterwarnings('ignore')
>>> def check(d, ok=True):
...   res = []
...   for k,v in d.items():
...     try:
...       z = dill.copy(v)
...       if ok: res.append(k)
...     except:
...       if not ok: res.append(k)
...   return res
... 
>>> check(failures)
[]
>>> check(registered, ok=False) # should be empty
['SymtableEntryType']
>>> check(succeeds, ok=False) # should be empty
['SymtableEntryType']
>>> 
>>> import builtins
>>> import types
>>> q = dill._dill._reverse_typemap
>>> p = {k:v for k,v in q.items() if k not in vars(builtins) and k not in vars(types)}
>>> set(p.keys()).difference(registered.keys())
set()
>>> set(registered.keys()).difference(p.keys())
set()
>>> 
```

Behavior is as expected, with the exception of `SymtableEntryType` -- which is not serializable

Issue #525, opened 2022-07-10
some errors with dump_module and load_module


mmckerns:
In trying out `dump_module` and `load_module` a bit more extensively after #507, and I ran across some interesting cases:

1) dump an imported builtin module that contains unpicklable objects
```
>>> import dill
>>> import math
>>> dill.dump_module('math.pkl', main=math)
>>> _math = dill.load_module('math.pkl')
>>> _math.sin(0)
0.0
>>> _math = dill.load_module('math.pkl', main='math')
>>> _math.sin(0)
0.0
>>> import abc
>>> dill.dump_module('abc.pkl', main=abc)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/mmckerns/lib/python3.7/site-packages/dill/_dill.py", line 488, in dump_module
    pickler.dump(main)
  File "/Users/mmckerns/lib/python3.7/site-packages/dill/_dill.py", line 880, in dump
    StockPickler.dump(self, obj)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/pickle.py", line 437, in dump
    self.save(obj)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/pickle.py", line 504, in save
    f(self, obj) # Call unbound method with explicit self
  File "/Users/mmckerns/lib/python3.7/site-packages/dill/_dill.py", line 2099, in save_module
    state=_main_dict)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/pickle.py", line 662, in save_reduce
    save(state)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/pickle.py", line 504, in save
    f(self, obj) # Call unbound method with explicit self
  File "/Users/mmckerns/lib/python3.7/site-packages/dill/_dill.py", line 1667, in save_module_dict
    StockPickler.save_dict(pickler, obj)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/pickle.py", line 859, in save_dict
    self._batch_setitems(obj.items())
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/pickle.py", line 885, in _batch_setitems
    save(v)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/pickle.py", line 504, in save
[...]

PR #527, opened 2022-07-16
Session: improvements to documentation + handle unpickleable objects

leogama:
**Edit: merged with #475.**

---

Additional changes to session/module saving and restoring based on discussions at #525.

- Created an `_open()` function that returns a file/stream context manager
- Created an `is_pickled_module()` function to test whether a pickle file was created by `dump_module()` and if it's module is importable or not. (It currently returns `True` for modules saved with `dill.dump(module, file)`.)
- Added a `refonfail` option for `dump_module()` to save objects that fail to pickle by reference, rolling back the pickle stream and saving one of the parent objects by reference if necessary.

### To-do

Documentation:

- [x] Explain the distinction between calling `dill.dump(module, file)` and `dill.dump_module(file, module)`, and the assumptions and use cases for each alternative — probably in the package's main documentation, or in a separate `session.py` module's docstring
- [x] Describe the effects of passing the `module` argument to `load_module()`
- [x] Better describe the side effects and the usage of the returned value from `load_module()`
- [x] Document the function `is_pickled_module(filename)` (what about its name and signature?)

Code:

- [x] Make `is_pickled_module()` return `False` for modules saved by name (and for modules saved by value inside other objects)
- [x] Deal with unpickleable objects in modules ~(likely in a separate PR)~
- [x] Deal with unpickleable variables in the module's namespace

@mmckerns ☝🏼 You may start drafting any of these if you feel inspired. Use the "suggested changes" in the review page or modify the code directly (adding new commits) if you are able to. I set the "allow editing by maintainers" option, although I don't know what it means...


leogama:
@mmckerns: [this docs page](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/committing-changes-to-a-pull-request-branch-created-from-a-fork) affirms you can push commits directly to the PR branch on my fork's repo. Have you ever tried this procedure?

mmckerns:
> Have you ever tried this procedure?

Yes.  I'm fine with that approach, or fork-PR'ing your fork. Either way.

leogama:
I may have successfully solved the problem with unpickleable objects. Added a `refonfail` option to `dump_session()`. I thought of calling it `refunpickleable`, but no one would get the name right. If you have any better idea...

I've overwritten the `save()` method to trap failed picklings and fallback to saving the object (or one of its parents) by reference. It's tricky to revert everything up to where the fallback strategy succeeds (must take care of framing, memo, trace state, etc.), but it seems to be working. I didn't test it extensively with the Stdlib modules yet.

Note: the framing trick probably has unhandled corner cases. It may be necessary to disable framing completely (by downgrading the protocol to a maximum of 3) when using this option.

mmckerns:
> It may be necessary to disable framing completely (by downgrading the protocol to a maximum of 3 or something else) when using this option.

I could see this definitely being the case, as the framing used in higher protocols is a good bit more complex.  I think the primary reason people will be using this functionality is to "save a session" -- hence the emphasis here should be on being able to save the session without failing... with pickle size a secondary concern (and pickling speed a distant third).

mmckerns:
> Document the function `is_module_pickle(filename)` (what about its name and signature?)

I'd probably prefer: `is_pickled_module(object)` or `is_pickled_module(file)`, where the former can take an object (including a file handle, or string file path) and the latter would expect a file handle or string file path.

leogama:
> > It may be necessary to disable framing completely (by downgrading the protocol to a maximum of 3 or something else) when using this option.
> 
> I could see this definitely being the case, as the framing used in higher protocols is a good bit more complex. I think the primary reason people will be using this functionality is to "save a session" -- hence the emphasis here should be on being able to save the session without failing... with pickle size a secondary concern (and pickling speed a distant third).

This isn't really about complexity. The problem is that the current frame of an object being pickled may be committed and flushed to file before an unpickleable child object raises an exception and can't even be saved by reference, forcing its siblings that were pickled previously with success to be discarded and one of its parents to be saved by reference.

leogama:
> I'd probably prefer: `is_pickled_module(object)` or `is_pickled_module(file)`, where the former can take an object (including a file handle, or string file path) and the latter would expect a file handle or string file path.

I've changed the name, but didn't understand the distinction you made. The parameter signature is identical to the those of the other `*_module()` functions.

leogama:
@mmckerns: I'm having difficulty with the ordering of module's members in the Sphinx output. Documentation says:

> #### Options
> `:members:`
>  (...)
>  For modules, __all__ will be respected when looking for members unless you give the ignore-module-all flag option. Without ignore-module-all, the order of the members will also be the order in __all__.
> (...)

And also:

> #### Configuration
> `autodoc_member_order`
> This value selects if automatically documented members are sorted alphabetical (value 'alphabetical'), by member type (value 'groupwise') or by source order (value 'bysource'). The default is alphabetical.
> (...)

I can't find the `ignorre-module-all` flag or the `autodoc_member_order` option set anywhere, but Sphinx doesn't respect the `__all__` ordering as it should. And the actual order is not even the supposed `alphabetical` default. It seems to be the `groupwise' one (classes first, then functions, etc.). For classes' members it uses `alphabetical` though.

mmckerns:
Is there a need for some particular ordering?

leogama:
> Is there a need for some particular ordering?

For modules, ideally something like:
- constants
- main functions/classes
- other functions/classes
- exceptions

I was thinking of putting the members in a logical sequence in the module `session`. For example:
1. `EXCLUDE` and `INCLUDE` constants
1. `dump_module`
1. `load_module`
1. `load_module_asdict`
1. `is_pickled_module`
1. `ModuleFilters`
1. `FilterRules` — base class of `ModuleFilters`
1. `FilterSet` — used in `FilterRules`

For class members, alphabetical order seems better. But I don't know if it's possible to use different orderings for modules and classes.

dill/_dill.py
```diff
@@ -410,9 +416,69 @@ def save_numpy_array(pickler, obj):
             raise PicklingError(msg)
         logger.trace_setup(self)
         StockPickler.dump(self, obj)
-
     dump.__doc__ = StockPickler.dump.__doc__
 
+    def save(self, obj, save_persistent_id=True, *, name=None):
+        """If self._refonfail is True, try to save object by reference if pickling fails."""
```
mmckerns, line 422: You shouldn't refer to `self` in the docs -- either use `refonfail is True` or `dill.settings['refonfail'] is True`... basically the doc should refer to how you'd expect a user to set it.  It's better than the `StockPickler.save` docs (i.e. None!), but this is a more complex signature, and I expect it a bit difficult to figure out what all the args do.  For example, I'm going to have to read the code to see what `name` does.  Also, you might want to be more specific when this function is called... is it after any time pickling fails, or just in certain cases (i.e. from certain function calls)?

dill/_dill.py
```diff
@@ -410,9 +416,69 @@ def save_numpy_array(pickler, obj):
...
+            # AttributeError may happen in the save_global() call from a child object.
+            if (type(error_stack) == AttributeError
+                    and "no attribute '__name__'" not in error_stack.args[0]):
+                raise
+            # Roll back the stream.
+            self._file_seek(position)
+            self._file_truncate()
+            # Roll back memo.
+            for _ in range(len(self.memo) - memo_size):
+                self.memo.popitem()  # LIFO order is guaranteed since 3.7
```
mmckerns, line 443: Have you tested this in the context of a `multiprocess.Pool` or a `multiprocess.dummy.Pool`?

dill/_dill.py
```diff
@@ -410,9 +416,69 @@ def save_numpy_array(pickler, obj):
...
+                    # Roll back trace state.
+                    self._trace_stack.pop()
+                    self._size_stack.pop()
+                raise error from error_stack
+            logger.trace(self, "# X: fallback to save_global: <%s object at %#012x>",
+                         type(obj).__name__, id(obj), obj=obj)
+
+    def _save_module_dict(self, obj):
+        """
+        Use object name in the module namespace as a last resource to try to
```
mmckerns, line 461: Did you mean: "last resort" or "last resource"?  Again, you might want to be more specific about "when pickling fails".

dill/_dill.py
```diff
@@ -410,9 +416,69 @@ def save_numpy_array(pickler, obj):
...
+        save it by reference when pickling fails.
+
+        Modified from Pickler.save_dict() and Pickler._batch_setitems().
```
mmckerns, line 464: This should be in a comment (not in the docstring), and may need a reference to the license (depending on what the license says).

dill/settings.py
```diff
@@ -19,7 +21,10 @@
     'fmode' : 0, #HANDLE_FMODE
     'recurse' : False,
     'ignore' : False,
+    'dump_module' : {
```
mmckerns, line 24: I don't really like the idea of a nested settings dict.  To me, the idea of `settings` is that it can be used to override the similarly named `kwargs` on certain pickling functions.  If `dump_module` is used as a key, why then not `dump` and `load`?  As long as the keys are unique, it shouldn't matter if it's a flat dict (the old `byref` <- `refimported` broke this).

dill/settings.py
```diff
@@ -19,7 +21,10 @@
...
+        'refimported': False,
+        'refonfail' : False,
```
mmckerns, line 26: Should the default be `True`?  I'm not suggesting it should be, but just want us to discuss it a bit.

dill/session.py
```diff
@@ -7,19 +7,42 @@
 # License: 3-clause BSD.  The full license text is available at:
 #  - https://github.com/uqfoundation/dill/blob/master/LICENSE
 """
-Pickle and restore the intepreter session.
+Pickle and restore the intepreter session or a module's state.
+
+The functions :py:func:`dump_module`, :py:func:`load_module` and
```
mmckerns, line 12: The doc here is good, but I think could use some rework with regard to the English. I'll come back to this.

dill/session.py
```diff
@@ -34,6 +57,54 @@
...
+            chunk = stream.read(n)
+            stream.seek(position)
+            return chunk
+        except (AttributeError, OSError):
+            raise NotImplementedError("stream is not peekable: %r", stream) from None
+
+def _open(file, mode, *, peekable=False):
+    """return a context manager with an opened file-like object"""
+    import io
+    attr = 'write' if 'w' in mode else 'read'
```
mmckerns, line 87: what about "write" file modes: `x`, `a`, and `r+`?

dill/session.py
```diff
@@ -149,6 +223,12 @@ def dump_module(
             similar but independent from ``dill.settings[`byref`]``, as
             ``refimported`` refers to virtually all imported objects, while
             ``byref`` only affects select objects.
+        refonfail: if `True`, objects that fail to pickle by value will try to
+            be saved by reference.  If this also fails, saving their parent
+            objects by reference will be attempted recursively.  In the worst
+            case scenario, the module itself may be saved by reference.  Note:
+            The file-like object must be seekable and truncable with this
```
mmckerns, line 230: truncatable

dill/session.py
```diff
@@ -213,32 +305,39 @@ def dump_module(
 
     from .settings import settings
     protocol = settings['protocol']
+    if refimported is None:
+        refimported = settings['dump_module']['refimported']
```
mmckerns, line 309: see my comment in `dill.settings` about a nested settings dict

dill/session.py
```diff
@@ -213,32 +305,39 @@ def dump_module(
...
         pickler._first_pass = True
-        pickler._main_modified = main is not pickler._original_main
+        if refonfail:
+            pickler._refonfail = True  # False by default
+            pickler._file_seek = getattr(file, 'seek', None)
+            pickler._file_truncate = getattr(file, 'truncate', None)
+            if hasattr(file, 'seekable') and not file.seekable():
+                pickler._file_seek = None
+            if pickler._file_seek is None or pickler._file_truncate is None:
+                raise TypeError("file must have 'tell', 'seek' and 'truncate'"
```
mmckerns, line 338: Hmm... I'm not sure if I like the doc here.  It's very accurate, however.

dill/session.py
```diff
@@ -247,64 +346,60 @@ def dump_session(filename=str(TEMPDIR/'session.pkl'), main=None, byref=False, **
...
     except (NotImplementedError, ValueError) as error:
         # ValueError occours when the end of the chunk is reached (without a STOP).
         if isinstance(error, NotImplementedError) and main is not None:
             # file is not peekable, but we have main.
             return None
-        raise UnpicklingError("unable to identify main module") from error
+        raise UnpicklingError("unable to identify module") from error
+
+def is_pickled_module(filename, importable: bool = True) -> bool:
+    """Check if a file is a module state pickle file.
```
mmckerns, line 383: It's not really transparent what `module state pickle file` means.  Your explanation in `Returns` is better.

dill/session.py
```diff
@@ -333,12 +428,66 @@ def load_module(
...
 
     Returns:
         A module object, if the saved module is not :py:mod:`__main__` or
-        a module instance wasn't provided with the argument ``main``.
+        a module instance wasn't provided with the argument ``module``.
+
+    Passing an argument to ``module`` forces `dill` to verify that the module
+    being loaded is compatible with the argument value.  Additionally, if the
+    argument is a module (instead of a module name), it supresses the return
+    value. Each case and behavior is exemplified below:
```
mmckerns, line 441: Thanks for detailed case-by-case examples. I'm going to have to come back to this.

mmckerns (changes requested):
Please address the review comments.  We can delay dealing with the two big blocks of doc that are noted as "come back to this later" until after everything else is dealt with.

dill/_dill.py
leogama: I can't even imagine how a pickler would be parallelized by using a `Pool`. The `memo` is an instance variable, and `memoize()` is not atomic. Calling `dump()` multiple times in parallel will probably fail (or generate corrupted pickles) even without these `refonfail`-related operations.

dill/_dill.py
leogama: I meant "last resort". There's not much specificity here... What I could add is

- When pickling an object raises an exception and the `save_global()` fallback call can't find the object by itself.

dill/_dill.py
leogama: I followed the Python License requirements and added the License Agreement in `LICENSE`, plus a copyright notice and a brief summary of the modifications to the code in the source file. Just pushed this, take a look.

dill/_dill.py
leogama: This is "private" API, it's an override of `StockPickler.save()`, which is not documented as you noted. Isn't it better to convert this to a comment? This function is **always** called, once for each object being pickled. And it always calls the superclass method, the only difference being that if `refonfail` is `True` it traps any common pickling exceptions and then it does its real job.

I wrote the last chunks of code in a rush, the documentation is full of gaps. Please keep pointing where the code is in need of minimal explanation. I'll add lines about the parameters.

dill/settings.py
leogama: I understand these parameter defaults in `settings` are general in spirit, i.e. they apply to all the main functions and methods where it makes sense to apply them (and exceptions, like `recurse` for `dump_module` are now documented). With these new settings there would be more exceptions than cases...

What about renaming the `dump_module` entry, to `session`? It matches the module name and would still make sense —except that someone may think that filters would somehow apply to `load_module()`. If a nested dictionary is the main problem, I suggest using `session.refimported`, `session.refonfail` and `session.filters` as keys directly in the `settings` dictionary.

dill/settings.py
leogama: Upsides:
- Less error messages in general
- Would lower the steepness of the learning curve and work out-of-the-box for most cases
- More people using this feature may unveil edge cases and let to a higher coverage of modules in the wild

Downsides:
- Would disable framing (speed optimization) by default, although I have no idea of the real impact 
- The particular state the user is interested in saving may not be saved (and they won't notice unless they activate and follow the tracing log or until they verify it worked after restoring the module)
- Currently, the whole module may end being saved by reference, i.e. no state saved at all...

(I'm thinking if this last one shouldn't raise a warning at least.)

I'm divided, but I think it's OK to have this option on by default.

dill/session.py
leogama: changed to `readonly = ('r' in mode and '+' not in mode)`

dill/session.py
leogama: I've never met this word before, in any language. I probably guessed it wrong with a Latin structure.

dill/session.py
leogama: It's a restriction of the implementation. I couldn't find better words... Even wrapping the stream in a `io.BufferedRandom` object wouldn't work. A plausible workaround for this would be to write the pickle stream to memory (`io.BytesIO`) and write to the file at the end. Except for really big objects and low memory situations, the overhead may be relatively small.

dill/_dill.py
mmckerns: something like this is a simple case of `dump` in parallel...
```
Python 3.7.13 (default, May 10 2022, 11:13:40) 
[Clang 10.0.1 (clang-1001.0.46.4)] on darwin
>>> 
>>> import multiprocess.dummy as threads
>>> p = threads.Pool()
>>> import dill
>>> y = p.map(dill.dumps, range(4))
>>> y
[b'\x80\x03K\x00.', b'\x80\x03K\x01.', b'\x80\x03K\x02.', b'\x80\x03K\x03.']
>>> 
```

LICENSE
```diff
@@ -33,3 +33,56 @@ WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR
 OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF
 ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+---------------------------------------------------------------------------
```
mmckerns, line 36: I think you shouldn't need to include this

dill/_dill.py
leogama: Well, in this example a new instance of `dill.Pickler` is created for each call to `dump()`. If it's this kind of thing you were concerned about, we are fine.

mmckerns: It would be worthwhile to run the `multiprocess` test suite to see if there are any new failures due to the change.  Typically, when something significant is touched, it's a good idea to run (at least) the `multiprocess`, `pathos`, `klepto`, and `mystic` test suites. They are a good stress test for `dill`, as they use `dill` as intended but have some "advanced pickling" cases.

dill/_dill.py
mmckerns: I know it's "private", and that it's always called... my point was that your doc is unclear.  I'd either change it to a comment (to be consistent with `pickle`) and/or improve the docs.

dill/_dill.py
mmckerns: > PSF's License Agreement and PSF's notice of copyright, i.e., "Copyright (c) 2001, 2002, 2003, 2004, 2005, 2006 Python Software Foundation; All Rights Reserved" are retained in Python alone or in any derivative version prepared by Licensee.

> In the event Licensee prepares a derivative work that is based on or incorporates Python or any part thereof, and wants to make the derivative work available to others as provided herein, then Licensee hereby agrees to include in any such work a brief summary of the changes made to Python.

It should be sufficient to summarize your changes, tag it with the copyright (right where you made the changes) and a link to the python license.

dill/settings.py
mmckerns: I like the "per module" idea. What about `dill.session.settings`, but not included in the `dill.settings` dict?

dill/settings.py
mmckerns: To me, it feels like the default should be `True`.  I expect that most people are going be interested in pickling unmodified modules, and "just want it to work".  So, if the majority of people will not use the default, then it's likely the wrong default. I'm fine with:

> the whole module may end being saved by reference

because, that is what `pickle` would give you.  I think it's also going to be expected that `dill` should do at least what `pickle` does, and using a default of `False` may give an impression that this is not the case.

If changing the default to `True`, we'd need to:

1. Document that `refonfail=False` may lead to faster serialization, but may fail when something in the module fails to serialize. 
2. Document that modifications will be ignored in any module that has to be stored by reference.

I think this is pretty acceptable, and a better choice.

dill/_dill.py
leogama: I extended/rewrote the documentation for save() and _save_module_dict().

dill/_dill.py
leogama: Just rewrote everything...

dill/_dill.py
leogama: Copyright law is hairy, but what I understood after reading a bit is that "derivative work" for code means copying, with or without changes, "major components" of the software. I believe that half a dozen lines of a function that must match the class API and the pickle protocol (and therefore can't be much different of what it is) don't constitute a major component. In this sense it wouldn't be necessary to add the copyright notice. But giving credit doesn't hurt.

dill/settings.py
leogama: @mmckerns Does my previous comment was appearing for you? It was in a "pending review" state. I don't know if that means it's like a draft and was visible just for me or if it was public.

LICENSE
```diff
@@ -33,3 +33,56 @@ WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR
...
+2. Subject to the terms and conditions of this License Agreement, PSF
+hereby grants Licensee a nonexclusive, royalty-free, world-wide
+license to reproduce, analyze, test, perform and/or display publicly,
+prepare derivative works, distribute, and otherwise use Python alone
+or in any derivative version, provided, however, that PSF's License
+Agreement and PSF's notice of copyright, i.e., "Copyright (c) 2001,
+2002, 2003, 2004, 2005, 2006, 2007, 2008, 2009, 2010, 2011, 2012,
+2013, 2014, 2015, 2016, 2017, 2018, 2019, 2020, 2021, 2022 Python
+Software Foundation; All Rights Reserved" are retained in Python alone
+or in any derivative version prepared by Licensee.
```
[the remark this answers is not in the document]
mmckerns: What I said is following the terms of the license agreement.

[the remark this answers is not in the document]
leogama: I wrote this two days ago... I think it wasn't visible for you. This is resolved.

dill/settings.py
leogama: Moved to `dill.session.settings`. Pending changes in #475

dill/settings.py
leogama: Everything but the item 2 was applied. But I'm not sure of what to do about objects in the module's namespace (not child objects) that fail to serialize. Consider these scenarios with two imported modules, `mod`  and `other_mod`, and an unpickleable variable named `var` that doesn't have an attribute `__name__` or `__qualname__` (note that there's no way to know where `var` was defined):

1.
    - `mod` defines `var`
    - `var` is not imported into other modules or is imported with aliases
    -  in `dump_module(module=mod)`, `var` may saved globally as `mod.var` or fail to save (e.g. for `__main__`)
2.
    - `mod` defines `var`
    - `other_mod` imports `var` from `mod`
    - in `dump_module(module=mod)`, `var` may saved globally as `mod.var` or `other_mod.var`
3.
    - `other_mod` defines `var`
    - `mod` imports `var` from `other_mod`
    - in `dump_module(module=mod)`, `var` may saved globally as `mod.var` or `other_mod.var`
4.
    - `other_mod` defines `var`
    - `mod` imports `var` as `var_alias`
    - in `dump_module(module=mod)`, `var` may saved globally as `mod.var_alias` or fail to save

**For each of these cases, which result is more appropriate?** Are all the desired behaviors compatible, i.e. is it possible to implement all of them consistently? Remember that, for the third scenario, `other_mod` may be a built-in module, which is safe, but may also be a third-party module not available in the unpickling environment. The behavior in this case can be stratified according to whether `other_mod` is a built-in module, is another module in the same package as `mod`, or is a third-party module.

dill/session.py
leogama: done

dill/session.py
leogama: Did basically this, with some modifications, and it worked! 🎉

dill/session.py
leogama: What about this for the short description:
> "Check if a file is a pickle file readable by `load_module()`."

mmckerns:
@leogama: looks like the latest merge of master left two `save` methods in `Pickler`.  Can you merge the save methods?

dill/__init__.py
```diff
@@ -68,6 +66,7 @@ def load_types(pickleable=True, unpickleable=True):
     Returns:
         None
     """
+    from importlib import reload
```
leogama, line 69: Where did this line come from?

dill/_dill.py
leogama: I run the test suites you suggested and `pathos`, `klepto` and `mystic` passed without errors. I've got an error with `multiprocess`, but it is identical to what I get with the `master` branch (something related to resources).

dill/settings.py
leogama: My previous comment brought up some situations where there's an unpickleable object in the namespace of the module being saved by `dump_module()`. If it has `__module__` and `__(qual)name__` attributes (5th scenario), it may be saved as global normally.  If that fails, then it falls into the same situation of an object without these attributes (1st to 4th scenarios).  I need to highlight that in these cases, the situations (2) and (3) are indistinguishable, there's no way to know where the object was originally defined.

Here, I'll further stratify the scenarios (1), (2) plus (3), and (4) presented earlier in a more practical way.  For the following discussion, a "builtin" module is a module for which `_is_builtin_module(module)` returns `True`, and a "user" module is a module for which the same test returns `False`.

Proposition: Assume an unpickleable object named `name` in the namespace of module `mod`. 

Possible cases:

1. `name` is only defined in `mod`
2. `name` is also defined in the "builtin" module `other`
3. object is defined in the "builtin" module `other` as `altname`
4. `name` is also defined in the "user" module `other` (except `__main__`)
5. object is defined in the "user" module `other` (except `__main__`) as `altname`

<table><thead>
  <tr>
    <th>Settings</th>
    <th colspan=2><code>refonfail=True</code> (default)</th>
    <th colspan=2><code>refimported=True</code></th>
    <th colspan=2>Both are <code>True</code></th>
    <th colspan=2>Both are <code>False</code></th>
  </tr>
  <tr>
    <th>Type of <code>mod</code></th>
    <th>"builtin"</th><th>"user"</th>
    <th>"builtin"</th><th>"user"</th>
    <th>"builtin"</th><th>"user"</th>
    <th>"builtin"</th><th>"user"</th>
  </tr>
</thead><tbody>
  <tr align="center">
    <td>Case (1)</td>
    <td>"mod.name"</td><td>exception</td>
    <td colspan=2>exception</td>
    <td>"mod.name"</td><td>exception</td>
    <td colspan=2>exception</td>
  </tr>
  <tr align="center">
    <td>Case (2)</td>
    <td colspan=2>"other.name"</td>
    <td colspan=2>"other.name"</td>
    <td colspan=2>"other.name"</td>
    <td colspan=2>exception</td>
  </tr>
  <tr align="center">
    <td>Case (3)</td>
    <td colspan=2>"other.altname"</td>
    <td colspan=2>"other.altname"</td>
    <td colspan=2>"other.altname"</td>
    <td colspan=2>exception</td>
  </tr>
  <tr align="center">
    <td>Case (4)</td>
    <td>"mod.name"</td><td>"other.name"</td>
    <td colspan=2>"other.name"</td>
    <td colspan=2>"other.name"</td>
    <td colspan=2>exception</td>
  </tr>
  <tr align="center">
    <td>Case (5)</td>
    <td>"mod.name"</td><td>"other.altname"</td>
    <td colspan=2>"other.altname"</td>
    <td colspan=2>"other.altname"</td>
    <td colspan=2>exception</td>
  </tr>
</tbody></table>

Rationale:

Focusing on the first two columns, which define the behavior of `refonfail`, there are two main cases: the module is a "builtin" module (e.g. `math` or installed `numpy`) or a "user" module (e.g. `__main__` or a module in the working directory).  And we have two main considerations to balance: how often should the operation fail (mainly concerning the default case) and wheter it's safe to consider that the object saved as global (without knowing if it actually belongs to the other module) will be there at unpickling.

### "builtin" modules

For "builtin" modules being saved, with the default options `refonfail=True` and `refimported=False`, `dump_module()` should work out-of-the-box, without errors.  As unpickleable objects in these modules are likely to always be defined, it's fairly safe to save them as references to the module itself (case 1, builtin).  Also if there's an object that is only found elsewhere in "user" modules, it should be saved in the same way (cases 4 and 5, builtin), as the most likely scenario is that such object was defined in the "builtin" module being saved and imported into the "user" modules (which have smaller chances to be available in the unpickling environment). Last, if the object is defined in another "builtin" module, it can be saved as a reference to that (cases 2 and 3, builtin).

### "user" modules

For "user" modules, which include `__main__`, if the unpickleable object is defined only in the module being saved, it's safer to assume the object may not be there when unpickling and therefore to raise an error (case 1, user).  While if the object is only defined in other "user" modules (except `__main__`), it may be fair to save it as a reference to one of them (cases 4 and 5, builtin), or alternatively to raise an error too.  This last case is the single one I'm not sure of how it should behave. Cases 2 and 3 should do the same as for "builtin" modules.

### Save the module by reference?

For "builtin" modules, unpickleable objects will always be saved by reference somehow (to a name in another module or to the module itself) with `refonfail=True`.  With `refonfail=False`, we can either save the module by reference, preferebly with a warning, or raise an error.  As this is not the default, I think that raising an error is fair, suggesting to the user the use of filters to exclude the offending object from the dump.

For "user" modules with default options, an unpickleable object will only raise an exception if it can't be found elsewhere (or only in `__main__`).  This case could force the module to be saved by reference and raise a warning.  For `refonfail=False`, I think it should always raise an error.

### Summary

The general behavior can be described as:

- For "builtin" modules, unpickleable objects are always saved by refence, even if it is a reference to the module itself.
- For "user" modules, it may raise an exception (internally) in some cases.
 
When a exception is raised:

- If `refonfail=True`, the default, save the module itself by reference with a warning.
- If `refonfail=False`, raise an error, so that the user has the opportunity to use exclusion filters to be able to save the state of other objects in the module.

I think it's no so hard to implement this logic as we already have `_is_builtin_module()`, `_locate_function()`, `_module_map()` and `_lookup_module()`. I'll try this tomorrow.

What do you think?

mmckerns: I think your proposal makes a lot of sense, and I have questions about the same cases as you have questions: (1) what to do for `refonfail=True` case 4/5 for "user", and (2) error vs warning behavior in some cases.

> While if the object is only defined in other "user" modules (except __main__), it may be fair to save it as a reference to one of them (cases 4 and 5, builtin), or alternatively to raise an error too. This last case is the single one I'm not sure of how it should behave.

I'd say, since it's `refonfail=True`... then this makes sense.  I'm wondering if there's a case where it fails in `mod` and is then saved by reference to `other.name` or similar... but the object is actually defined in `mod`, and `other.name` is just a reference to the original in `mod`.  The question is: is it possible for this to go wrong... where save-by-reference to `other.name` is the wrong thing to do.  I think we need to try to think up failure scenarios for this case.

> With refonfail=False, we can either save the module by reference, preferebly with a warning, or raise an error. As this is not the default, I think that raising an error is fair, suggesting to the user the use of filters to exclude the offending object from the dump.

I think I agree here.

> For "user" modules with default options, an unpickleable object will only raise an exception if it can't be found elsewhere (or only in __main__). This case could force the module to be saved by reference and raise a warning.

See my first comment.

> For refonfail=False, I think it should always raise an error.

I agree here.

Generally, I think a very solid and logical proposal.  I may need to construct some test cases and explore how they fail or succeed... this should better inform the cases we are unsure about.

dill/session.py
mmckerns: I like "Check if the pickle file was generates with `dump_module`".  You, could then, if you like, also add "and thus can be loaded with `load_module`".

dill/settings.py
leogama: Good that you agreed with most of my reasoning. I'll answer to your considerations later today.

While I'm advancing with the draft implementation of this logic, I'd like to propose a hierarchy of "safeness" for modules to be referred to when saving by reference. If an object is found in multiple modules, the safest alternative would always be preferred, as it has more chances to be there at unpickling (it also applies to `refimported`).

The proposed hierarchy is:
1. Standard Library modules
2. modules of the same package as the module being saved
3. installed modules in general ("builtin" modules)
4. non-installed modules ("user" modules)
5. `__main__` — never safe

Cases 1 to 3 would fall under the "builtin" umbrella of the table from my previous comment.

mmckerns: I'm not sure what you mean by "safeness" or "safe".  Is it just more granularity to "user" vs "builtin"?

leogama: "safer" == "greater probability of the module being available in the unpickling environment and of the object being defined in it (and pertain to it)"

For example, consider a module named `pack.mod`. At loading its state with `load_module()`, we can assume that both Standard Library modules and other modules from the package `pack` are available (otherwise, `pack.mod` itself wouldn't be loadable). So, if at saving its state with `dump_module()`, the unpickleable object `pack.mod.obj` can be saved globally either as `std.lib.obj` or as `pack.other.obj`, should we prefer one over the other? I argue that yes, because the most likely scenario is that the object is defined in the Standard Library module `std.lib` and imported from it in both `pack.mod` and `pack.other`. Plus, there's a small chance that, in the unpickling environment, the module `pack.mod` may exist but `pack.other` may not (for example, if we had slightly different versions of the package). For StdLib packages, however, that's highly unlikely.

That's the logic behind the whole hierarchy. It's an empirical way of maximizing the probability of success of saving by reference in the module's namespace.

This is almost orthogonal to the stratification of self/"builtin"/"user" behavior from that table. The only change is that modules that are part of the same package as the module being saved (if the module is part of any package) are put together with the "builtin" modules even if the package is not appropriately installed. Basically, this hierarchy will define which reference to another module is better, but not if that should be used instead of a reference to the module itself, or if an exception should be raised.

mmckerns:
I'm closing this, as it's been merged with #475.

dill/settings.py
leogama: (I likely wrote this a week ago but didn't press "send"...)

"safer" == "greater probability of the module being available in the unpickling environment and of the object being defined in it (and pertain to it)"

For example, consider a module named `pack.mod`. At loading its state with `load_module()`, we can assume that both Standard Library modules and other modules from the package `pack` are available (otherwise, `pack.mod` itself wouldn't be loadable). So, if at saving its state with `dump_module()`, the unpickleable object `pack.mod.obj` can be saved globally either as `std.lib.obj` or as `pack.other.obj`, should we prefer one over the other? I argue that yes, because the most likely scenario is that the object is defined in the Standard Library module `std.lib` and imported from it in both `pack.mod` and `pack.other`. Plus, there's a small chance that, in the unpickling environment, the module `pack.mod` may exist but `pack.other` may not (for example, if we had slightly different versions of the package). For StdLib packages, however, that's highly unlikely.

That's the logic behind the whole hierarchy. It's an empirical way of maximizing the probability of success of saving by reference in the module's namespace.

This is almost orthogonal to the stratification of self/"builtin"/"user" behavior from that table. The only change is that modules that are part of the same package as the module being saved (if the module is part of any package) are put together with the "builtin" modules even if the package is not appropriately installed. Basically, this hierarchy will define which reference to another module is better, but not if that should be used instead of a reference to the module itself, or if an exception should be raised.

leogama: Answering your previous comment:

> > While if the object is only defined in other "user" modules (except **main**), it may be fair to save it as a reference to one of them (cases 4 and 5, builtin), or alternatively to raise an error too. This last case is the single one I'm not sure of how it should behave.
> 
> I'd say, since it's `refonfail=True`... then this makes sense. I'm wondering if there's a case where it fails in `mod` and is then saved by reference to `other.name` or similar... but the object is actually defined in `mod`, and `other.name` is just a reference to the original in `mod`. The question is: is it possible for this to go wrong... where save-by-reference to `other.name` is the wrong thing to do. I think we need to try to think up failure scenarios for this case.

This problem of saving a reference to an object of other module that we don't know where was defined is the same that already existed for the `refimported` option (the old `byref` for modules). It could go wrong in many cases, yes, although I think they are a small minority of cases.

What I could think of to minimize this was: get the module associated with the object. It may be `obj.__module__` or `type(obj).__module__`. If it matches the module being pickled, don't save it as a reference to other modules, as it's most likely that the object was defined in the module itself. Secondly, if the module being pickled is part of a package, and the module associated with the object is from this package, only save it as a reference to a module of the same package.

The reason for the second rule is that packages in general are independent and have a unidirectional relationship, i.e. one acts as a library and the other as a client. So, suppose we are pickling a module `pack1.bar`, and we hit an unpickleable object of type `pack1.baz.Class`. We then find the same object also defined in a module `pack2.foo`. The most probable scenario is that `pack2` uses `pack1` as a library and `pack2.foo` imported the object from `pack1.bar` (the module we are pickling), not the opposite. The alternative hypothesis is that `pack2.foo` used `pack1.baz.Class` to create the object and then injected it in the `pack1.bar` namespace, which is very unlikely.

> Generally, I think a very solid and logical proposal. I may need to construct some test cases and explore how they fail or succeed... this should better inform the cases we are unsure about.

Well, I implemented tests to check the stratified behavior and the hierarchy of preferences when there are multiple choices of references to use.

Do you want some tests that would fail? I can think of a couple ones.

mmckerns: When a package has some fraction of a billion users, the thing you have to remember is that any time you think it's a rare corner case, someone will be attempting that rare corner case.  So, then the assessment is: what is the cost if `dill` serializes something incorrectly? Or what happens later to users with stored pickle files if the behavior is then corrected?  An at first seemingly rare, but actually somewhat common, is your latter example but the package is so large it has sub-packages (so like `numpy`, `sklearn`, `scipy`, etc), so you'd have `pack.pack2.foo` using `pack.pack1.baz.Class` to create an object and inject it into the `pack.pack1.bar`... which is suddenly a lot less unlikely.

leogama: Sorry, I wasn't clear. I did notice that while implementing these package-related logic. With "package" I meant "top-level package", not the subpackage (so `module.__package__.partition('.')[0]` when there's a package value) . I tested it on all modules loaded by default by `numpy`, `pandas`, `scipy`, `sklearn` and also `IPython`, and this value was consistent.

Look, if you are not 100% confortable with this feature, maybe it's better to leave it off by default in this release, see how it's used in the wild and if bugs/problems come up and then consider enabling it by default in a future release if we feel more confident.

A different thing that I thought of was to have a more granular control of for which modules these settings (`refonfail` and `refimported`) are enabled. We could have more options, something like (thinking of a bit field):
- 0 or False: disabled
- 1 or True: enabled for all modules
- 2: enabled for Standard Library modules
- 4: enabled for installed packages
- 8: enabled for user modules
- 16: enabled for __main__
- 32: enabled for module-type objects

Usage would be similar to the `os` flags for `open()`, like:
```python
import dill
from dill.session import settings
settings['refonfail'] = dill.MOD_STDLIB | dill.MOD_INSTALLED
```

mmckerns: > Sorry, I wasn't clear. I did notice that while implementing these package-related logic. With "package" I meant "top-level package", not the subpackage (so `module.__package__.partition('.')[0]` when there's a package value). I tested it on all modules loaded by default by `numpy`, `pandas`, `scipy`, `sklearn` and also `IPython`, and this value was consistent.

Excellent.  Thanks for the clarification.  Also, your idea about granular control is interesting. I don't think it should be addressed right now, however.  Post-release.

leogama: Right, I agree. I would only add this granular control now to have as default the `refonfail` option enabled only for Standard Library modules (and maybe for installed user modules). If we'll keep the current default, then leave that for later.

leogama:
@mmckerns Let's reopen this one. I've updated the PR from the current state of #475

PR #529, opened 2022-07-22
Fix pickling errors thrown when saving some Stdlib modules

leogama:
I'll add some small pickling errors' fixes here while I test the "save unpickleable objects by reference" option.

### DIFF
--- a/dill/_dill.py
+++ b/dill/_dill.py
@@ -2237,7 +2237,9 @@ def save_type(pickler, obj, postproc_list=None):
            #print (_dict)
            #print ("%s\n%s" % (type(obj), obj.__name__))
            #print ("%s\n%s" % (obj.__bases__, obj.__dict__))
-            for name in _dict.get('__slots__', ()):
+            slots = _dict.get('__slots__', ())
+            if type(slots) == str: slots = (slots,) # __slots__ accepts a single string
+            for name in slots:
                 del _dict[name]
             _dict.pop('__dict__', None)
             _dict.pop('__weakref__', None)
--- a/dill/_dill.py
+++ b/dill/_dill.py
@@ -1555,13 +1555,6 @@ def _get_attr(self, name):
     # stop recursive pickling
     return getattr(self, name, None) or getattr(__builtin__, name)
 
-def _dict_from_dictproxy(dictproxy):
-    _dict = dictproxy.copy() # convert dictproxy to dict
-    _dict.pop('__dict__', None)
-    _dict.pop('__weakref__', None)
-    _dict.pop('__prepare__', None)
-    return _dict
-
 def _import_module(import_name, safe=False):
     try:
         if import_name.startswith('__runtime__.'):
@@ -2238,20 +2231,17 @@ def save_type(pickler, obj, postproc_list=None):
         obj_recursive = id(obj) in getattr(pickler, '_postproc', ())
         incorrectly_named = not _locate_function(obj, pickler)
         if not _byref and not obj_recursive and incorrectly_named: # not a function, but the name was held over
-            if issubclass(type(obj), type):
-                # thanks to Tom Stepleton pointing out pickler._session unneeded
-                _t = 'T2'
-                logger.trace(pickler, "%s: %s", _t, obj)
-                _dict = _dict_from_dictproxy(obj.__dict__)
-            else:
-                _t = 'T3'
-                logger.trace(pickler, "%s: %s", _t, obj)
-                _dict = obj.__dict__
+            # thanks to Tom Stepleton pointing out pickler._session unneeded
+            logger.trace(pickler, "T2: %s", obj)
+            _dict = obj.__dict__.copy() # convert dictproxy to dict
            #print (_dict)
            #print ("%s\n%s" % (type(obj), obj.__name__))
            #print ("%s\n%s" % (obj.__bases__, obj.__dict__))
-            for name in _dict.get("__slots__", []):
+            for name in _dict.get('__slots__', ()):
                 del _dict[name]
+            _dict.pop('__dict__', None)
+            _dict.pop('__weakref__', None)
+            _dict.pop('__prepare__', None)
             if obj_name != obj.__name__:
                 if postproc_list is None:
                     postproc_list = []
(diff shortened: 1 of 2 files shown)


leogama:
@mmckerns: I've been working with module-saving code for months and didn't realize until now that the function `_load_module()` doesn't do exactly what its name suggests. For example, I was thinking of pickling the `settings` submodule to preserve `dump_module()` options and filters set between a series of python invocations using an automating tool. However:

```python
>>> import dill                                                                                                                                                                                
>>> dill.dump_module(module="dill.settings")                                                                                                                                                   
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-2-84d05d12f876> in <module>
----> 1 dill.dump_module(module='dill.settings')

~/Projects/forks/dill/dill/session.py in dump_module(filename, module, refimported, refonfail, **kwds)
    297         main = _import_module(main)
    298     if not isinstance(main, ModuleType):
--> 299         raise TypeError("%r is not a module" % main)
    300     original_main = main
    301     if refimported:

TypeError: {'protocol': 4, 'byref': False, 'fmode': 0, 'recurse': False, 'ignore': False, 'dump_module': {'refimported': False, 'refonfail': False}} is not a module
```

In this case, the `dill.settings` object is actually the `sys.modules['dill.settings'].settings` dictionary imported in the module `dill`. That's why I passed `"dill.settings"` as the `module` argument instead of `dill.settings`. But, inside `dump_module()`, `_import_module("dill.settings")` returns the dictionary, not the module. I don't know how much it is by design though.

It seems that `_import_module()` works much more like `Unpickler.find_class()` than like `importlib.import_module()`, while I expected pretty much the opposite (i.e. that in my example it would try to return the submodule `dill.settings` and, if that failed, it would fallback to return the attribute `settings` of the module `dill`).

I'm really confused...

---

By the way, I'm testing the "`refonfail`" option of `dump_module()` to save objects that fail to pickle by reference and I'm happy to announce that **dill is currently able to pickle 99% of the Standard Library**'s modules and public submodules (the ones listed in docs.python.org). Most of the fails are due to infinite recursion, which I shall investigate soon.

mmckerns:
> dill is currently able to pickle 99% of the Standard Library's modules and public submodules 

That's great news.

> It seems that` _import_module()` works much more like `Unpickler.find_class()` than like `importlib.import_module()`

Unfortunately there are a lot of undocumented non-public functions in `dill`.  However, you can see from the code that it does import the module (or submodule), so it's an appropriate name... it's just not the `dill` equivalent of `importlib.import_module()`.

leogama:
> > It seems that` _import_module()` works much more like `Unpickler.find_class()` than like `importlib.import_module()`
> 
> Unfortunately there are a lot of undocumented non-public functions in `dill`. However, you can see from the code that it does import the module (or submodule), so it's an appropriate name... it's just not the `dill` equivalent of `importlib.import_module()`.

Yes, it seems that there are places where using `importlib.import_module()` is more appropriate than `_import_module()`.

But take a look at this listing:

```python
>>> import dill
>>> dill.__version__
'0.3.5.1'
>>> import importlib
>>> dill_settings = importlib.import_module('dill.settings')
>>> dill_settings
<module 'dill.settings' from '/home/leogama/.local/lib/python3.8/site-packages/dill/settings.py'>
>>> dill.copy(dill_settings)
{'protocol': 4, 'byref': False, 'fmode': 0, 'recurse': False, 'ignore': False}
>>> import pickletools
>>> pickletools.dis(pickletools.optimize(dill.dumps(dill_settings, protocol=0)))
    0: c    GLOBAL     'dill._dill _import_module'
   27: (    MARK
   28: V        UNICODE    'dill.settings'
   43: t        TUPLE      (MARK at 27)
   44: R    REDUCE
   45: .    STOP
highest protocol among opcodes = 0
```

---

By the way, could you help me with another issue real quick? I'd like to add push a commit to `master` just moving the session-related stuff to a new `session.py` submodule. I did that in both the #475 and #527 branches, but this code movement is preventing me from reviewing changes in this segment as git treats this as brand new file...

dill/_dill.py
```diff
@@ -1555,13 +1555,6 @@ def _get_attr(self, name):
     # stop recursive pickling
     return getattr(self, name, None) or getattr(__builtin__, name)
 
-def _dict_from_dictproxy(dictproxy):
```
mmckerns, line 1558: I think this is ok to delete as it's not a registered function (so no issues with stored pickles requiring it).

dill/_dill.py
```diff
@@ -2238,28 +2231,27 @@ def save_type(pickler, obj, postproc_list=None):
...
+            _dict.pop('__prepare__', None)
             if obj_name != obj.__name__:
                 if postproc_list is None:
                     postproc_list = []
                 postproc_list.append((setattr, (obj, '__qualname__', obj_name)))
             _save_with_postproc(pickler, (_create_type, (
                 type(obj), obj.__name__, obj.__bases__, _dict
             )), obj=obj, postproc_list=postproc_list)
-            logger.trace(pickler, "# %s", _t)
+            logger.trace(pickler, "# T2")
```
mmckerns, line 2254: I guess it's also ok we don't have a "T3" case any longer.

dill/_dill.py
```diff
@@ -2238,28 +2231,27 @@ def save_type(pickler, obj, postproc_list=None):
...
-                _t = 'T2'
-                logger.trace(pickler, "%s: %s", _t, obj)
-                _dict = _dict_from_dictproxy(obj.__dict__)
-            else:
-                _t = 'T3'
-                logger.trace(pickler, "%s: %s", _t, obj)
-                _dict = obj.__dict__
+            # thanks to Tom Stepleton pointing out pickler._session unneeded
+            logger.trace(pickler, "T2: %s", obj)
+            _dict = obj.__dict__.copy() # convert dictproxy to dict
```
mmckerns, line 2236: You've added a few new operations to the old "T3" version, so that should be slightly slower... but the way a dict copies, it should be negligible.

mmckerns (approved):
LGTM

dill/_dill.py
leogama: Isn't "T3" a special case for old-style classes from Python 2?

mmckerns: ...hmm, now that you mention it, I think so.

PR #486, opened 2022-05-26
Lookup __qualname__ instead of __name__ in Python 3

anivegesana:
Piece of code I forgot to open a PR for in dill 0.3.5

https://github.com/uqfoundation/dill/issues/482#issuecomment-1133791230

Fixes #411

Fixes #482

Fixes #483, fixes bitranox/wrapt_timeout_decorator#41

PR #534, opened 2022-08-04
Shared namespaces

### LINKED ISSUE #466: Function namespace preservation
Continuing the discussion from https://github.com/uqfoundation/dill/issues/462#issuecomment-1112761901

Even cloudpickle's behavior is a little bit finicky and has some unintuitive behavior. I think that there are design decisions that need to be made with regards to how we handle global dictionaries and functions.

```py
import cloudpickle

def f():
    global x
    x = 5

def g():
    global x
    x = 0
    f()
    assert x == 5


G, GG = cloudpickle.loads(cloudpickle.dumps((g, g.__globals__)))

G.__globals__ != GG # Seems odd
```

### LINKED ISSUE #532: __globals__ of function in deleted module restored inconsistently
My application sometimes pickles functions after the module they were imported from has been removed from `sys.modules` (for reasons we hopefully don't have to get into). Usually `dill` handles this just fine, but there seems to be a bug when the pickled function is used in a closure which also maintains a reference to the pickled function's `__globals__` dictionary. The following code yields an assertion failure with `dill` 0.3.5.1 (and also the latest version on `master`; in each case I was using Python 3.9):

```
import dill
import sys

import other    # another module; see below

def make():
    func = other.blah
    globs = func.__globals__
    def inner():
        assert globs is func.__globals__
    return inner

f = make()

del sys.modules['other']

f()     # assertion passes here
d = dill.dumps(f)
f2 = dill.loads(d)
f2()    # assertion fails here
```
where `other.py` contains the code:
```
def blah():
    return 42
```
The assertion fails because in the unpickled version of `inner`, the free variable `func` is bound to the unpickled version of `blah` but the free variable `globs` is bound to a dict which is *not* the globals dict of `func` (nor is it the globals dict of `other.blah`, in fact).

What seems to be happening is that when `blah` is unpickled, `dill` calls `_create_function` with the argument `fglobals` set to an empty dictionary, so that the line
```
func = FunctionType(fcode, fglobals or dict(), fname, fdefaults, fclosure)
```
creates a *new* dict to be the globals dict of the new function. Then when the "post-processing" updates the cells of `inner`, it sets the cell for `globs` to the *old* dictionary instead of this newly-created one. To confirm this diagnosis, I removed the `or dict()` from the line above, and changed the "F1" case of `save_function` to always set `globs = globs_copy` (i.e. removing the `else` branch where it sets `globs = {'__name__': obj.__module__}`), and indeed the problem went away. I'm not sure how to properly fix the issue, though.

anivegesana:
I'll fix anything tomorrow night. It is late.

Fixes #532, fixes #466

### DIFF
--- a/dill/_dill.py
+++ b/dill/_dill.py
@@ -366,6 +366,7 @@ class Pickler(StockPickler):
         self._recurse = settings['recurse'] if _recurse is None else _recurse
         self._postproc = OrderedDict()
         self._file = file
+        self._globals_cache = {}
 
     def dump(self, obj): #NOTE: if settings change, need to update attributes
         # register if the object is a numpy ufunc
@@ -1792,17 +1793,14 @@ def save_function(pickler, obj):
         _postproc = getattr(pickler, '_postproc', None)
         _main_modified = getattr(pickler, '_main_modified', None)
         _original_main = getattr(pickler, '_original_main', __builtin__)#'None'
+        _globals_cache = getattr(pickler, '_globals_cache', None)
         postproc_list = []
+
+        globs = None
         if _recurse:
             # recurse to get all globals referred to by obj
             from .detect import globalvars
             globs_copy = globalvars(obj, recurse=True, builtin=True)
-
-            # Add the name of the module to the globs dictionary to prevent
-            # the duplication of the dictionary. Pickle the unpopulated
-            # globals dictionary and set the remaining items after the function
-            # is created to correctly handle recursion.
-            globs = {'__name__': obj.__module__}
         else:
             globs_copy = obj.__globals__
 
@@ -1815,6 +1813,18 @@ def save_function(pickler, obj):
             elif globs_copy is not None and obj.__module__ is not None and \
                     getattr(_import_module(obj.__module__, True), '__dict__', None) is globs_copy:
                 globs = globs_copy
+
+        if globs is None:
+            # Add the name of the module to the globs dictionary and prevent
+            # the duplication of the dictionary. Pickle the unpopulated
+            # globals dictionary and set the remaining items after the function
+            # is created to correctly handle recursion.
+            if _globals_cache is not None and obj.__globals__ is not None:
+                if id(obj.__globals__) not in _globals_cache:
+                    globs = {'__name__': obj.__module__}
+                    _globals_cache[id(obj.__globals__)] = globs
+                else:
+                    globs = _globals_cache[id(obj.__globals__)]
             else:
                 globs = {'__name__': obj.__module__}
 
--- a/dill/tests/_globals_dummy.py
+++ b/dill/tests/_globals_dummy.py
@@ -1,0 +1,10 @@
+# This file is used by test_shared_globals in test_functions.py
+
+x = 3
+
+def h():
+   print(x)
+
+def g():
+   h()
+
--- a/dill/tests/test_functions.py
+++ b/dill/tests/test_functions.py
@@ -132,7 +132,27 @@ def test_code_object():
         except Exception as error:
             raise Exception("failed to construct code object with format version {}".format(version)) from error
 
+def test_shared_globals():
+	import dill, _globals_dummy as f, sys
+
+	for recurse in False, True:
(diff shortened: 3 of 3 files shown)


anivegesana:
I remember why I didn't do this earlier. It makes `save_module_dict` more complicated, but we need to make it simpler to do #486.

mmckerns:
When this is ready, either mark it as ready for review, or request a review.

leogama:
> I still don't have permissions to do that, but this PR is now ready

This is the GitHub workflow we kind of discovered and have being using:
1. I open a **Draft** Pull Request
2. We discuss some ideia, I push some commits
3. After a while, when I reach the point where review is necessary, I change the PR from Draft to **Ready for review**
4. _mmckerns_ receives a notification (I think) and assigns himself as a reviewer
5. _mmckerns_ does the review and either merges, or leaves comments and requests changes (and I'm notified)
6. I apply the requested and other changes
7. As a review was already done, I'm able to **re-request a review** for _mmckerns_
8. _mmckerns_ receives a notification
9. `GOTO` (5)

(If you didn't open the PR in draft mode, you may still convert it to a draft.)

> @mmckerns @leogama Look at the comment concerning cPickle first, because I do not know if that has higher priority.

I'll take a look at this later today.

leogama:
@anivegesana I read your changes more thoroughly and couldn't find any problem relative `dump_module()`. As the attributes `_session` and `_recurse` are mutually exclusive, the chances of unexpected behaviors are small. But you left some indentation tabs in `test_functions.py`.

About conflicts with the other PR, I'll have to solve it there.

leogama:
@mmckerns and @anivegesana: I'm thinking of ways to reorganize and simplify segment of `save_function()` below. For that I need to know if there is any situation where a function's `__globals__` or a module's `__dict__` is `None`. Is that possible at all?

https://github.com/uqfoundation/dill/blob/7fe9bb55d48fde49e296d8d0f1aecc48a3ff1387/dill/_dill.py#L1825-L1871

anivegesana:
Just test if `globs is None`. `globs_copy` is a list of items to update the globals dictionary cache.

leogama:
I've opened a PR in your fork: https://github.com/anivegesana/dill/pull/2

dill/_dill.py
```diff
@@ -1198,12 +1200,37 @@ def save_module_dict(pickler, obj):
...
+        logger.trace(pickler, "# D5")
     else:
         logger.trace(pickler, "D2: %s", _repr_dict(obj)) # obj
-        if is_dill(pickler, child=False) and pickler._session:
+        if _is_dill and pickler._session:
             # we only care about session the first pass thru
             pickler._first_pass = False
-        StockPickler.save_dict(pickler, obj)
+
+        # IMPORTANT: update the following code whenever save_dict is changed in pickle.py
```
mmckerns, line 1219: It appears that the code for `save_dict` in all supported version of python are is the same (at least from `if pickler.bin` to `pickler.memoize(obj)`). The idea here is to insert the code with `__name__` to `save_dict` while not yielding control to the `StockPickler`, correct?  It seems that `save_dict` hasn't changed as far back as at least 3.1, so I'm not worried about the copy being made here.

dill/tests/_globals_dummy.py
```diff
@@ -0,0 +1,10 @@
+# This file is used by test_shared_globals in test_functions.py
```
mmckerns, line 1: I think it should be possible to use or modify a simple test like `test_functors` so that file is used as opposed to using an additional file like this one.

mmckerns (commented):
Looks fine to me.  I'll do some testing and also have to understand what concerns @leogama has with your solution aside from there's some intersection with code in some of his PRs. Then I can move this to approve/request.

EDIT: Appears to correct behavior seen in #532, however doesn't seem to affect code initiating #466.

anivegesana:
I will need to look into this again at some point since it was a while that I looked at this before and forgot what these PRs were for. If I remember correctly, @leogama's changes were supposed to update the PRs to work with the code on master.

PR #536, opened 2022-08-08
Alternative fix for bug when pickling function from unloaded module (fixes #532)

### LINKED ISSUE #532: __globals__ of function in deleted module restored inconsistently

leogama:
This patch modifies `_locate_function()` and `save_function()` to allow dill to save a **function** from a module that was unloaded, i.e. deleted from the `sys.modules` dictionary, by reference (if it would be saved as global had the module not been unloaded). Note: this is independent of the shared namespace problem.

It may be possible to do the same for classes and methods, but these cases are more complex.

### DIFF
--- a/dill/_dill.py
+++ b/dill/_dill.py
@@ -60,7 +60,7 @@ import gc
 # import zlib
 from weakref import ReferenceType, ProxyType, CallableProxyType
 from collections import OrderedDict
-from functools import partial
+from functools import lru_cache, partial
 from operator import itemgetter, attrgetter
 GENERATOR_FAIL = False
 import importlib.machinery
@@ -1028,6 +1028,23 @@ def _import_module(import_name, safe=False):
             return None
         raise
 
+@lru_cache(maxsize=2**3)
+def _import_module_copy(module_name, safe=False):
+    """load (a copy of) a module without adding it to sys.modules"""
+    import importlib.util
+    try:
+        spec = importlib.util.find_spec(module_name)
+        if spec is None:
+            raise ModuleNotFoundError("No module named %r" % module_name)
+        module = importlib.util.module_from_spec(spec)
+        spec.loader.exec_module(module)
+    except ImportError:
+        if safe:
+            return None
+        else:
+            raise
+    return module
+
 # https://github.com/python/cpython/blob/a8912a0f8d9eba6d502c37d522221f9933e976db/Lib/pickle.py#L322-L333
 def _getattribute(obj, name):
     for subpath in name.split('.'):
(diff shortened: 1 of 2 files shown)


anivegesana:
I am confused about the purpose of caching the namespace. I have seen situations similar to #532 in popular packages ([torch.package](https://pytorch.org/docs/stable/package.html)) but never for standard library modules, probably because you can't change them, so there is no way to reload them.

Plus, what if someone changes a file and calls `importlib.reload` to forcibly reload a package with new source code. The cached namespace wouldn't be updated and would cause confusion. Is there any reason that this should be included? If so, there should be a way to turn it off.

leogama:
> I am confused about the purpose of caching the namespace. I have seen situations similar to #532 in popular packages ([torch.package](https://pytorch.org/docs/stable/package.html)) but never for standard library modules, probably because you shouldn't change them, so there is no reason to reload them in the first place.

I used a stdlib module for demonstration, but this mechanism would be triggered mainly by third-party packages, as in the example you mentioned. Reloaded modules are not covered by this patch, only "unloaded" ones. With a reloaded module, modified or not, dill (pickle) would still raise an "object is not the same as XXX.YYY" error.

> Plus, what if someone changes a file and calls `importlib.reload` to forcibly reload a package with new source code. The cached namespace wouldn't be updated and would cause confusion.

This conflict would only happen with a series of events like:
```python
import importlib
import sys
import dill
from module import func
dill.dumps(func)  # save as global
del sys.modules['module']
dill.dumps(func)  # add 'module' to _modules_copy's cache, save as global (this is the difference in behavior)
#--- Modify func in module.py ---#
import module  # re-import into sys.modules (creates a new module instance)
importlib.reload(module)  # required to update module.func?
dill.dumps(func)  # func is not module.func, raise error
del sys.modules['module']
dill.dumps(func)  # func is different from cached _modules_copy('module').func, save by value
```

In this last statement, the functions wouldn't match in the `_locate_function()` test and the logic would fall back to the old behavior (the function would be saved by value). The namespaces are cached because they are copies independent of the loaded modules and so that a new copy of the unloaded module isn't created for every object. Not caching would matter only in that case of "load -> import object -> unload -> pickle object -> reload with changes -> unload -> pickle object". The cache could be associated with the `Pickler` instance, although it wouldn't have effect for a sequence of calls to `dill.dump[s]` then.

**Note:** I have no preference for this or any of the other proposed solutions, I'm mostly throwing things against the wall.

Issue #540, opened 2022-08-10
Pickling class instance in dict fails unless you try pickling the object directly first


JohnGoertz:
I've found a very bizarre issue where, if I try to pickle an object (with various instance attributes) in a dictionary using dill, it fails on the first attempt. When un-pickling the object, it appears none of the instance attributes have been pickled along with the object. No matter how many times I re-run that cell (in a jupyter notebook), it fails. If I then try to pickle the same object directly, it works, and a subsequent attempt at pickling it in a dictionary now works too.

I've attached reproducible code below, and you can run it in [this Colab notebook](https://colab.research.google.com/drive/1Jdq_K-gyuDKwTcRdY8L7bK5AmLf_aJKn?usp=sharing).

Note, I am the author of gumbi, so please feel free to let me know if this is some issue with gumbi itself!

## Imports and setup
```python
!pip install gumbi

import dill
import numpy as np
import gumbi as gmb

stdzr = gmb.Standardizer(d={'μ': -0.307, 'σ': 0.158}, log_vars=['d'])
pa = gmb.parray(d=np.arange(5,10)/10, stdzr=stdzr)
pa
# ('d',): [(0.5,) (0.6,) (0.7,) (0.8,) (0.9,)]
```

## First attempt
Saving the object in a dictionary fails, the instance attributes seemingly don't get pickled:
```python
with open('test.pkl', 'wb') as f:
    dill.dump({'pa': pa}, f)
    
with open('test.pkl', 'rb') as f:
    loaded = dill.load(f)
    
loaded['pa']
# AttributeError: 'ParameterArray' object has no attribute 'names'
```

## A diversion
Saving the object on its own works:
```python
with open('test.pkl', 'wb') as f:
    dill.dump(pa, f)
    
with open('test.pkl', 'rb') as f:
    loaded = dill.load(f)
    
loaded
# ('d',): [(0.5,) (0.6,) (0.7,) (0.8,) (0.9,)]
```

## Second attempt
And now saving the object in a dictionary works:
```python
with open('test.pkl', 'wb') as f:
    dill.dump({'pa': pa}, f)
    
with open('test.pkl', 'rb') as f:
    loaded = dill.load(f)
    
loaded['pa']
# ('d',): [(0.5,) (0.6,) (0.7,) (0.8,) (0.9,)]
```

Note that running the "diversion" before ever running the dictionary pickle works too!

mmckerns:
[...]

PR #542, opened 2022-08-11
Optimize the numpy hook

leogama:
As this is now called from `Pickler.save()`, which can be called from a dozen to thousands of times for a single "dump":
- Moved repeated logic to outside the numpy test functions
- Only create a save function once for each type
- Remove logic related to Python 2 old style classes
- Substituted black magic like comparing `id(obj1.__bound_method__) == id(obj2.__bound_method__)`* by  `type(obj1).__unbound_method__ is type(obj2).__unbound_method__)`

(*) I was surprised that it worked at all, must be some kind of optimization but this is not described in the language.

Amend to #540

### DIFF
--- a/dill/_dill.py
+++ b/dill/_dill.py
@@ -91,51 +91,39 @@ def __hook__():
     from numpy import dtype as NumpyDType
     return True
 if NumpyArrayType: # then has numpy
-    def ndarraysubclassinstance(obj):
-        if type(obj) in (TypeType, ClassType):
-            return False # all classes return False
-        try: # check if is ndarray, and elif is subclass of ndarray
-            cls = getattr(obj, '__class__', None)
-            if cls is None: return False
-            elif cls is TypeType: return False
-            elif 'numpy.ndarray' not in str(getattr(cls, 'mro', int.mro)()):
+    def ndarraysubclassinstance(obj_type):
+        try: # check if is ndarray or subclass of ndarray
+            if all((c.__module__, c.__name__) != ('numpy', 'ndarray') for c in obj_type.__mro__):
                 return False
-        except ReferenceError: return False # handle 'R3' weakref in 3.x
-        except TypeError: return False
+        except (ReferenceError, TypeError): # TypeError is (was) for what?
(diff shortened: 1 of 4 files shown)


leogama:
@mmckerns: The tests are failing because `CLibraryLoaderType` (`ctypes.cdll` of type `ctypes.LibraryLoader`) is registered as unpickleable for CPython < 3.11b3, but it pickles just fine. This class has `__reduce__()` and `__reduce_ex__()` methods implemented.

https://github.com/uqfoundation/dill/blob/bf0f4fa48358fac0a9246a3f12575767d1d14b6f/dill/_objects.py#L191-L193

mmckerns:
Obviously, the types should just be registered the once, so yeah checking if they are in the dispatch table is the right thing to do.  Thanks for the PR.

mmckerns:
> The tests are failing because CLibraryLoaderType (ctypes.cdll of type ctypes.LibraryLoader) is registered as unpickleable

It was failing due to `getattr(cdll, 'mro', int.mro)` throwing a `OSError` in the `numpy*` check functions. Fixed in: 43fd2ca082448e545ca98a4ba72327a0912e7672

leogama:
> It was failing due to `getattr(cdll, 'mro', int.mro)` throwing a `OSError` in the `numpy*` check functions.

This doesn't make sense... The old versions of the `numpy*` functions were effectively calling `getattr(cdll.__class__, 'mro', int.mro)` (on the `cdll` class, not on the object). Furthermore, the lines that were generating the error are removed in this PR. So if the supposed failure to pickle `cdll` has anything to do with that functions, just the adjustment to `_objects.py` should be enough to solve the test error.

mmckerns:
> Note: on my machine, with all Python versions, cdll pickles just fine, so I can't track down this issue.

I have VMs for all supported platforms. The `OSError` didn't occur on linux, but did otherwise... so, yes, once that was adjusted for, then `cdll` pickled... and adjusting `_objects.py` enables the tests to pass.

mmckerns (approved):

PR #555, opened 2022-10-04
New auxiliary function to get value for options: `_getopt()`

leogama:
From #475

### DIFF
--- a/dill/_dill.py
+++ b/dill/_dill.py
@@ -229,10 +229,9 @@ def dump(obj, file, protocol=None, byref=None, fmode=None, recurse=None, **kwds)
     See :func:`dumps` for keyword arguments.
     """
     from .settings import settings
-    protocol = settings['protocol'] if protocol is None else int(protocol)
-    _kwds = kwds.copy()
-    _kwds.update(dict(byref=byref, fmode=fmode, recurse=recurse))
-    Pickler(file, protocol, **_kwds).dump(obj)
+    protocol = int(_getopt(settings, 'protocol', protocol))
+    kwds.update(byref=byref, fmode=fmode, recurse=recurse)
+    Pickler(file, protocol, **kwds).dump(obj)
     return
 
 def dumps(obj, protocol=None, byref=None, fmode=None, recurse=None, **kwds):#, strictio=None):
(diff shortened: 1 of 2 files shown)


dill/_dill.py
```diff
@@ -317,6 +316,26 @@ class PicklingWarning(PickleWarning, PicklingError):
...
+        self._byref = _getopt(settings, 'byref', kwds=kwds)
+    """
+    # Sanity check, it's a bug in calling code if False.
+    assert kwds is None or arg is None
+    if kwds is not None:
+        arg = kwds.pop(key, None)
+    if arg is not None:
+        return arg
+    else:
+        return settings[key]
```
mmckerns, line 337: It's a bit unorthodox that the alternate is given as the first two values, and then the arg/kwd you are primarily interested in is given after that.  Consider the syntax of `getattr` and similar.

So, this function pops `key` out of `kwds` (isn't get a better choice?), and if `key` is found and not `None`, then it is returned.  If not found, then use `arg` if `arg` is provided, unless it's `None`... then default back to `settings[key]`.  It's also expected that `arg` and `kwds` are never both given... so essentially, you have two one-liners in a single function, where any interaction between the two one-liners is an error.  I'm not seeing the utility of this function, in that case.  I could see using it, maybe, if you wanted to hide the import of `settings` within the function -- so `_getopt(key, arg=None, *, kwds=None)`.

dill/_dill.py
```diff
@@ -317,6 +316,26 @@ class PicklingWarning(PickleWarning, PicklingError):
...
+    Examples:
+
+        # With an explicitly named argument:
+        protocol = int(_getopt(settings, 'protocol', protocol))
+
+        # With a named argument in **kwds:
+        self._byref = _getopt(settings, 'byref', kwds=kwds)
+    """
+    # Sanity check, it's a bug in calling code if False.
+    assert kwds is None or arg is None
```
mmckerns, line 331: To me, an `AssertionError` is not expected... shouldn't it be a `ValueError`?

mmckerns (commented):
I'm not sure if this is a good idea in its current form.

mmckerns (changes requested):
Please respond to the review comments.

PR #556, opened 2022-10-04
New `is_pickled_module()` function

leogama:
From #475 / #527
Also fixes `_identify_module()` -> raises exception for modules saved with `dump()`


mmckerns (changes requested):
Looks really nice. Sorry about the delay.  Let's have a small think about naming of the function, what it returns, and the naming of `importable` kwarg (and it's docs). Otherwise LGTM.

PR #450, opened 2022-01-27
Support ABC and Enums

anivegesana:
1) Support ABC classes (supersedes #427)
2) Support Enum classes
3) Replace `_dict_from_dictproxy` with cloudpickle equivalent that doesn't save the members of the parent classes

Fixes #250, fixes #332, fixes #365

PR #558, opened 2022-10-13
Drop _abc._abc_data from saving. 

ryanthompson591:
This prevents an exception when havng a class implementing an ABCMeta class.

Please review this carefully to make sure it fits with dills workflow.

Background:
When pickling a class that inherits from metaclass=abc.ABCMeta dill crashes with an error:
```
TypeError: cannot pickle '_abc._abc_data' object
```

This can be bypassed by dropping _abc_impl from the dictionary in save_type.

### DIFF
--- a/dill/_dill.py
+++ b/dill/_dill.py
@@ -29,6 +29,7 @@ import warnings
 from .logger import adapter as logger
 from .logger import trace as _trace
 
+import abc
 import os
 import sys
 diff = None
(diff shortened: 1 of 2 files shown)


mmckerns:
@ryanthompson591: Thanks for submitting this PR. This seems to be a duplicate of a subset of #450. Note also your PR is currently failing for `pypy`.

ryanthompson591:
It looks like that PR does something similar to what I am doing, on line 2006 of _dill.py in their PR.

@mmckerns - what do you suggest.  Should I drop this PR since it might duplicate work already being done?

mmckerns:
> Should I drop this PR since it might duplicate work already being done?

I'm pretty sure it can be left to the other PR, and we can close this one.

mmckerns:
closing

PR #598, opened 2023-06-06
Fixed crashed in _dill._is_builtin_module when a module's __file__ is None

miguelinux314:
## Summary

The `_is_builtin_module` function of `dill/_dill.py` has the following check in line 1587

`if not hasattr(module, "__file__"): return True`

If the check is not passed, then `os.path.realpath(module.__file__)` is called later in that function (line 1591).

This causes dill to crash when a module has `__file__` set to `None`. This PR fixes that check so that when `module.__file__ is None`, the module is recognized as not builtin.

Notes:
* Even though https://docs.python.org/3/reference/import.html#file__ specifies that when `__file__` is set it is a string, it will actually be none when importing a module without an `__init__.py` file in it.
* This PR only changes the behavior of dill to prevent TypeError crashes. Code that did not crash before is unaffected by this PR.
* No issue created prior to this PR)

### DIFF
--- a/dill/_dill.py
+++ b/dill/_dill.py
@@ -1612,6 +1612,7 @@ def save_weakproxy(pickler, obj):
 
 def _is_builtin_module(module):
     if not hasattr(module, "__file__"): return True
+    if module.__file__ is None: return False
     # If a module file name starts with prefix, it should be a builtin
     # module, so should always be pickled as a reference.
     names = ["base_prefix", "base_exec_prefix", "exec_prefix", "prefix", "real_prefix"]


miguelinux314:
Thank you @mmckerns for the quick reply.

Yes, the `__file__` attribute of any module (built-in or not) can be *manually* set to anything. 

On the other hand, under certain conditions (see https://github.com/python/cpython/blob/423459be2f0b6d007e5f235f39d80044cb099faf/Lib/importlib/_bootstrap.py#L667, caused by https://bugs.python.org/issue32305) `importlib` does set it to `None` without complaining or manually fidgeting with the `__file__` attribute.

Based on this, I don't think one can get the `__file__` attribute to differ both from a valid path string and `None` without actively *trying* to break the code. That's why I checked only for None and left other cases to crash. However, if you feel the dill library would be sturdier considering the manual fiddling cases, my recommendation would be a try-catch of lines 1620 to 1629 with `TypeError`. I'd be happy with either solution.

Thanks again for you time and help

mmckerns:
I'll take it as is. thanks

PR #613, opened 2023-08-19
test for qualname in get_typedef_type

### LINKED ISSUE #612: Class attributes identical to inherited ones are wrongly deleted
In 0.3.7, attributes in the `__dict__` of a class are now deleted by `dill` if they have the same name and value as an attribute of a superclass. For example:

```python
import dill

class Foo:
    x = 1

class Bar(Foo):
    x = 1  # identical value as superclass

assert 'x' in Bar.__dict__  # passes

data = dill.dumps(Bar)
Bar2 = dill.loads(data)

assert 'x' in Bar2.__dict__  # fails with dill 0.3.7 (works with 0.3.6)
```

This behavior is caused by the `_get_typedict_type` function added in #577. Based on its docstring it seems the idea was to exclude inherited methods, but the code actually excludes any attribute whose value is identical to the inherited value. This is not safe (arguably even in the case of methods) because as the code above demonstrates, it changes the behavior of code which inspects the `__dict__` directly. Obviously the above code is a toy example, but we have real code which does this (basically to set a non-inherited flag on classes) and which was broken by `dill` 0.3.7.

In case it matters, I'm using Python 3.11.1 on macOS; we've also seen the problem under Python 3.10.12 on Ubuntu.

mmckerns:
## Summary
Fixes #612. Test if values in the inherited dict are the same only if they have a qualname. If qualname is not used, then `'__module__'` is incorrectly set to `'dill._dill'` when it is `'__main__'`... and additionally, attributes defined in the super class with values that are flyweights or singletons (i.e. have the same hash) that are also defined in the derived class should always appear in the derived class's `__dict__`.

This PR also changes an assertion in test_abc that synchronizes previously incongruous behavior of ABC classes (there was a TODO note). I believe the new behavior is correct, in that is is more consistent with other classes behavior after pickling an instance of the class.

## Checklist
**Documentation and Tests**
- [x] Artifacts produced with the main branch work as expected under this PR.

**Release Management**
- [ ] Requested a review.

### DIFF
--- a/dill/_dill.py
+++ b/dill/_dill.py
@@ -1693,7 +1693,7 @@ def _get_typedict_type(cls, clsdict, attrs, postproc_list):
     for name, value in dict.items(clsdict):
         try:
             base_value = inherited_dict[name]
-            if value is base_value:
+            if value is base_value and hasattr(value, '__qualname__'):
                 to_remove.append(name)
         except KeyError:
             pass
--- a/dill/tests/test_abc.py
+++ b/dill/tests/test_abc.py
@@ -77,7 +77,8 @@ def test_abc_non_local():
     # Set a property that StockPickle can't preserve
     instance.bar = lambda x: x**2
     depickled = dill.copy(instance)
-    assert type(depickled) is not type(instance)
+    assert type(depickled) is type(instance) #NOTE: issue #612, test_abc_local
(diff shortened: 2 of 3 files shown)


mmckerns:
@anivegesana: you may want to check if this has any backward incompatible breakage for ABC classes of pickles of old ABC classes.

Issue #628, opened 2023-11-02
`save_function()` can't save function in a submodule that has the same name as an attribute of the parent module


kelvinburke:
Hi,
So if there is a module `test_module/__init__.py`:
```
from .test import *
```
and then in `test_module/test.py` there is:
```
def test():
    pass
def test_function():
    pass
```
Then if you `import test_module` then the name `test_module.test` goes to the function `test_module.test()` instead of the submodule `test_module.test`.
If we then try to (dill) pickle either one of the functions it will raise an error:
```
import dill
import tempfile
import test_module
file = tempfile.TemporaryFile()
dill._dill.StockPickler(file).save(test_module.test_function)
# OR
dill._dill.StockPickler(file).save(test_module.test)
# OR
import pickle
pickle._Pickler(file).save(test_module.test_function)
```
Any of the above 3 will throw an error:
```
Traceback (most recent call last):
  File "C:\code\test_bug3.py", line 5, in <module>
    dill._dill.StockPickler(file).save(test_module.test_function)
  File "C:\...\AppData\Local\Programs\Python\Python311\Lib\pickle.py", line 560, in save
    f(self, obj)  # Call unbound method with explicit self
    ^^^^^^^^^^^^
  File "C:\...\.venv\Lib\site-packages\dill\_dill.py", line 1940, in save_function
    for stack_element in _postproc:
TypeError: 'NoneType' object is not iterable
```

This is using `dill==0.3.7` and Python 3.11 and on the master branch too https://github.com/uqfoundation/dill/commit/f66ed3b602d6e9b942542491614dfd18622583d7, and a repo to reproduce it is: https://github.com/kelvinburke/dill-issue

I think this is because the function `_import_module()` returns the function `test_module.test` instead of the submodule of the same name.
I think this can be easily fixed with a check that the `getattr(__import__(module, None, None, [obj]), obj)` returns the right type
See commit: https://github.com/kelvinburke/dill/commit/228a7006f603acb4b441b32710b7b5dbbbfcc301

Note I think this is the same problem causing https://github.com/uqfoundation/dill/issues/604 but this a slightly different error and simpler to reproduce.

I will open a PR soon that I think will fix both.

mmckerns:
I can reproduce the error.  However, you'll note that if you are using the `StockPickler`, you are essentially using the `Pickler` from `pickle` and not `dill`.  If you use the intended `dill.Pickler`, it seems to work as expected.
```
[...]

Issue #604, opened 2023-07-13
RecursionError with function that has the same name as its owning module


pauldmccarthy:
Howdy, I am experiencing a `RecursionError` with dill 0.3.6 and the master branch (82431115). It seems to be related to the structure of my code, where I have a function which is located within a module of the same name. For example, let's say I have the following code structure:

```
pkg/
    __init__.py:
        from pkg.somefunc import somefunc
        from pkg.somemod  import modfunc

    somefunc.py:
        def somefunc():
            print('pkg.somefunc.somefunc')

    somemod.py:
        def modfunc():
            print('pkg.somemod.modfunc')
```

I think that the issue is to do with the `dill._dill._import_module` (which is used by the `dill._dill._locate_function` function) - with the above example code structure, these functions produce the following results (I have commented out the `__all__` declaration at the top of `_dill.py`):

```python
import dill._dill as _dill
import pkg

print(_dill._import_module('pkg.somemod'))
print(_dill._locate_function(pkg.modfunc))
print(_dill._import_module('pkg.somefunc'))
print(_dill._locate_function(pkg.somefunc))
```

```
<module 'pkg.somemod' from '/.../pkg/somemod.py'>
True
<function pkg.somefunc.somefunc()>
False
```

The logic within the `_import_module` function seems to assume that, if given a path to a module `pkg.somefunc`, then that module must be accessible from within the namespace of the parent package `pkg` - specifically [this line](https://github.com/uqfoundation/dill/blob/824311158af3c73551141b4445ad21e7bafd363f/dill/_dill.py#L1028) resolves to  `getattr(__import__('pkg'), 'somefunc')`, which in my case will return the `pkg.somefunc.somefunc` function rather than the `pkg.somefunc` module.

Triggering the `RecursionError` is not trivial - the simple example above is not actually enough to trigger the error (there must be some short-circuiting logic somewhere), and I don't know `dill` (or pickling in general) well enough to come up with a minimal example. 

However, I have prepared a small reproducible example based on the codebase I was working in when I came across this issue at https://github.com/pauldmccarthy/dill-issue

PR #629, opened 2023-11-02
Import submodule properly when there is an attribute of the module with the same name

kelvinburke:
## Summary
Fixes https://github.com/uqfoundation/dill/issues/628 and fixes https://github.com/uqfoundation/dill/issues/604 : I have checked that they don't throw errors now.

Not sure if it breaks anything else - some of the tests on my machine failed on the master branch - but the tests run the exact same after this change.

Not sure if a test should be added, something like https://github.com/kelvinburke/dill-issue - but I'm not sure the best way?

## Checklist

**Documentation and Tests**
- [ ] Added relevant tests that run with `python tests/__main__.py`, and pass.
- [ ] Artifacts produced with the main branch work as expected under this PR. -- Not sure what this means?

**Release Management**
- [ ] Added rationale for any breakage of backwards compatibility. -- Not sure if this does break backwards compatibility?
- [x] Requested a review.

Let me know what you think!

### DIFF
--- a/dill/_dill.py
+++ b/dill/_dill.py
@@ -1023,9 +1023,12 @@ def _import_module(import_name, safe=False):
             items = import_name.split('.')
             module = '.'.join(items[:-1])
             obj = items[-1]
+            submodule = getattr(__import__(module, None, None, [obj]), obj)
+            if isinstance(submodule, (ModuleType, type)):
+                return submodule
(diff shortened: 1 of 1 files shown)


kelvinburke:
I'm also not sure this is the best way to do it, or if there are other types that are returned by `__import__` other than `ModuleType, type` ?

PR #651, opened 2024-04-02
fix fencepost error when getting source inside decorator in interpreter (fixes #603)

### LINKED ISSUE #603: dill.source.getsource returns partial source when called inside a decorator in a REPL
I noticed this behaviour  - I'm able to easily work around in my case, but noting it here.

In a REPL, Python 3.11, when getsource is invoked inside a decorator function (so trying to get the source of the function that is being decorated right now), partial source is returned  (rather than the whole source, or an error). This problem does not happen with the same code in a Python file executed from the command line.

```
>>> import demo
>>> @demo.d
... def m():
...   print("hi")
...   return 7
... 
@demo.d
def m():
  print("hi")

>>> 
```

but in this case, with getsource running much later (getsource is called when the decorated function is invoked), I get the full source:

```
>>> import demo
>>> @demo.e
... def m():
...   print("hi")
...   return 8
... 
>>> m()
@demo.e
def m():
  print("hi")
  return 8
```

with demo.py:

```
import dill.source

def d(f):
    print(dill.source.getsource(f))
    return f

def e(f):
    def wrapped():
      print(dill.source.getsource(f))
    return wrapped
```

timkpaine:
## Summary
I *believe* this fixes an off-by-one error for functions defined in the interpreter.

I am not sure how to test this as it only runs when directly instantiated in an interpreter.
Fixes https://github.com/uqfoundation/dill/issues/603.

Inline test:

```python
from dill.source import getsource

def my_decorator(foo):
    s = getsource(foo)
    print(s)
    assert s.endswith("return 1\n")

@my_decorator
def foo():
  return 1
```

**Before**

```raw
@my_decorator
def foo():

  File "<stdin>", line 4, in my_decorator
AssertionError
```

**After**
```python
@my_decorator
def foo():
    return 1
>>>
```

Please let me know any suggestions for how to implement a test for this!

## Checklist

**Documentation and Tests**
- [ ] Added relevant tests that run with `python tests/__main__.py`, and pass.
- [ ] Added relevant documentation that builds in sphinx without error.
- [ ] Added new features that are documented with examples.
- [ ] Artifacts produced with the main branch work as expected under this PR.

**Release Management**
- [ ] Added "Fixes #NNN" in the PR body, referencing the issue (#NNN) it closes.
- [ ] Added a comment to issue #NNN, linking back to this PR.
- [ ] Added rationale for any breakage of backwards compatibility.
- [ ] Requested a review.

### DIFF
--- a/dill/source.py
+++ b/dill/source.py
@@ -150,7 +150,7 @@ def findsource(object):
         if err:
             raise IOError(err)
         lbuf = readline.get_current_history_length()
-        lines = [readline.get_history_item(i)+'\n' for i in range(1,lbuf)]
+        lines = [readline.get_history_item(i)+'\n' for i in range(1,lbuf+1)]
     else:
         try: # special handling for class instances
             if not isclass(object) and isclass(type(object)): # __class__


mmckerns:
thanks for the patch

Issue #243, opened 2017-11-01
instance doesn't use pickled __class__ when same-named class exists


piccolbo:
I was just reading @mmckerns answer on [SO](https://stackoverflow.com/questions/25613543/how-to-use-dill-to-serialize-a-class-definition/25623532#25623532) and marveling at the power of Dill. So I tried his first example and to my surprise it fails exactly like pickle would, whereas @mmckerns states "Pickle would blow up on the above." Which I think implies "dill doesn't". Then I found out that it fails in iPython but not python. Does anyone know what the cause is? Are there known limitations in using with iPython?

```
In [59]: import dill
    ...: 
    ...: class Foo(object):
    ...:   def bar(self, x):
    ...:     return x+self.y       
    ...:   y = 1
    ...: 
    ...: f = Foo()
    ...: _Foo = dill.dumps(Foo)
    ...: _f = dill.dumps(f)
    ...: class Foo(object):
    ...:   def bar(self, x):
    ...:     return x*self.z  
    ...:   z = -1
    ...: f_ = dill.loads(_f)
    ...: f_.y
    ...: 
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-59-de94877455f9> in <module>()
     14   z = -1
     15 f_ = dill.loads(_f)
---> 16 f_.y

AttributeError: 'Foo' object has no attribute 'y'

```

I reported this on [SO](https://stackoverflow.com/questions/25613543/how-to-use-dill-to-serialize-a-class-definition/25623532#25623532) as a comment, then I thought that thread was a bit too past its prime for anyone to pay attention, so I am reporting here and cross linking.

mmckerns:
Still haven't gotten to this, but I just wanted to make it more clear what is going on.  This behavior works as expected from python, but in IPython it doesn't.  I'll have to investigate what IPython is doing that is causing the stored `__class__` to be ignored.

mmckerns:
So, this works:
```
In [1]: import dill

In [2]: class Foo(object):
   ...:     def bar(self, x):
   ...:         return x+self.y
   ...:     y = 1
   ...: 

In [3]: f = Foo()

In [4]: _Foo = dill.dumps(Foo)

In [5]: _f = dill.dumps(f)

In [6]: del Foo

In [7]: del f

In [8]: f_ = dill.loads(_f)

In [9]: f_
Out[9]: <__main__.Foo at 0x1046bb7f0>

In [10]: f_.__class__
[...]

PR #704, opened 2025-02-27
fix: get correct source from redefined class

### LINKED ISSUE #703: getsource not get correct source code
```
import dill


class A:
    def a(self):
        print("a")


print(dill.source.getsource(A))


class A:
    def b(self):
        print("b")


print(dill.source.getsource(A))
```
Which will return
```
class A:
    def b(self):
        print("b")

class A:
    def b(self):
        print("b")
```

Seraphli:
## Summary

Fixes #703 

## Checklist

**Documentation and Tests**

**Release Management**
- [x] Requested a review.

### DIFF
--- a/dill/source.py
+++ b/dill/source.py
@@ -26,7 +26,7 @@ import linecache
 import re
 from inspect import (getblock, getfile, getmodule, getsourcefile, indentsize,
                      isbuiltin, isclass, iscode, isframe, isfunction, ismethod,
-                     ismodule, istraceback)
+                     ismodule, istraceback, currentframe)
 from tokenize import TokenError
 
 from ._dill import IS_IPYTHON
@@ -101,6 +101,19 @@ def _matchlambda(func, line):
         return True
     return False
 
+def get_outer_frame():
+    """
+    Get the outermost frame that is not inside source.py.
+    Returns the frame and its line number.
+    """
+    frame = currentframe()
+    if not frame: return None, None
+    source_file = __file__
+    while frame:
+        if frame.f_code.co_filename != source_file:
+            return frame, frame.f_lineno
+        frame = frame.f_back
+    return None, None
 
 def findsource(object):
     """Return the entire source file and starting line number for an object.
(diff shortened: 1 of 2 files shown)


Seraphli:
I use current main branch code to run tests on Python 3.9.9 on my local linux machine and get the errors. That confirms these errors are not introduced by this PR.

mmckerns:
Here's a bit of a out-of the box thought regarding this PR.  What if you were to pickle the class, then get the source from the pickled instance? `dill` isolates the (class/function) context, and pickles all of that... so identifying the correct instance could at the very least be confirmed by what get pickled.

Seraphli:
I don’t quite understand the scenario you're describing. Do you have any test cases? What exactly do you mean by 'get the source from the pickled instance'? I've never performed such an operation before, so I'm not entirely clear on the situation you're referring to.

Seraphli:
I find when wrap `getsource` in a function will cause a new problem. I fix that problem in the new commit.

mmckerns:
See: https://github.com/uqfoundation/dill/issues/243 for an example of serializing class definition with the instance

Seraphli:
I looked into the issue you mentioned and tested a small example:  
```python
import dill

class Foo(object):
    y = 1

_Foo_bits = dill.dumps(Foo)

class Foo(object):
    z = 1

_Foo = dill.loads(_Foo_bits)
print(dill.source.getsource(_Foo))
print(_Foo.y)
```
I observed that the current code returns:  
```  
class Foo(object):
    z = 1
1  
```  
Are you expecting it to return:  
```  
class Foo(object):
    y = 1
1  
```  
instead?

Seraphli:
```python
import dill
import inspect


class Foo(object):
    def foo(self):
        return self.x

    x = 1


class Foo(object):
    def bar(self):
        return self.y

    y = 1


_Foo_bits = dill.dumps(Foo)


class Foo(object):
    def foo(self):
        return self.z

    z = 1


_Foo = dill.loads(_Foo_bits)
print(dill.source.getsource(_Foo))
print(inspect.getsource(_Foo))
print(_Foo.y)
f = _Foo()
print(f.bar())
```
The result is
```python
class Foo(object):
    def foo(self):
        return self.z

    z = 1

class Foo(object):
    def foo(self):
        return self.x

    x = 1

1
1
```
I guess in this situation, the only solution is to save the code when pickling the class object. I can't think of any other approaches.

There might be another complex solution - we could first locate the invocation points and analyze the source file's AST to trace the origin of class variables. However, this would inevitably lead to another issue: if the variable is imported from other modules, there's no way to trace back the variable's origin just analyzing one file's AST.

mmckerns:
> instead?

What I was alluding to from that issue is that `dill` serializes the class definition and you can control whether you use the serialized class definition or not with a `setting` (keyword) on `load`.  However, `get_source` always uses the buffer... and I'm wondering if, instead of trying the buffer first, it could first inspect the serialized object. The serialized object has all of the necessary code and dependencies contained.

Seraphli:
Is there a way to extract the code from a serialized class object? I didn't find any API in `dill`. Anyway in my case I use a decorator to modify the code of one class using AST, then return and replace the orignal class definition. So  I need `dill.getsource` when python goes through class definitions. If there is a way to extract the code from a serialized class object in current main branch, I would like to use that. That can solve my case.

mmckerns:
No, there isn't a way to do it currently, or I would have pointed you at it.
If you use `dill.source.getsource(obj)`, where `obj` is a serialized object, you get this: `"import dill\ndill.loads(...pickled...)"` where `...pickled...` is the pickled string.

PR #744, opened 2026-02-06
Fix saving __qualname__ of functions

### LINKED ISSUE #602: Metadata of wrapper functions not preserved
`dill` seems to not preserve the metadata of functions wrapped with `functools.wraps`. With `dill` 0.3.6 on Python 3.11.1, the following code:
```python
import functools

import dill

def a(x):
    return 42

@functools.wraps(a)
def b(x):
    return a(x)

data = dill.dumps(b)
c = dill.loads(data)

print(f'{a=}, {b=}, {c=}')
```
prints something along the lines of
```
a=<function a at 0x108164a40>, b=<function a at 0x1081963e0>, c=<function b at 0x108ca2fc0>
```
where the unpickled function should have name `a` like the original does.

I thought the issue might be that the `__qualname__` attribute of the function wasn't getting saved, but setting that attribute manually rather than using `functools.wraps` actually works; the code

```python
def d(x):
    return 42

d.__qualname__ = 'foo'

data = dill.dumps(d)
e = dill.loads(data)

print(f'{d=}, {e=}')
```
prints something like
```
d=<function foo at 0x10bd58860>, e=<function foo at 0x10c892fc0>
```
where now the name of the unpickled function is correct. So I don't know what's going on.

dfremont:
## Summary

This PR fixes the saving of the `__qualname__` attribute of functions when it is the same as `__name__` but different from `__code__.co_qualname` (a situation that arises when using `functools.wraps`: see issue #602 for details).

Fixes #602. This changes existing behavior, so is technically not backwards-compatible, but I think the existing behavior was clearly wrong. (It was also inconsistent between module-level and local functions.)

## Checklist

**Documentation and Tests**

**Release Management**
- [x] Requested a review.

### DIFF
--- a/dill/_dill.py
+++ b/dill/_dill.py
@@ -2031,7 +2031,7 @@ def save_function(pickler, obj):
             fattr = getattr(obj, fattrname, None)
             if fattr is not None:
                 state_dict[fattrname] = fattr
-        if obj.__qualname__ != obj.__code__.co_qualname:
+        if obj.__qualname__ != getattr(obj.__code__, "co_qualname", obj.__code__.co_name):
             state_dict['__qualname__'] = obj.__qualname__
         if '__name__' not in globs or obj.__module__ != globs['__name__']:
             state_dict['__module__'] = obj.__module__
--- a/dill/_dill.py
+++ b/dill/_dill.py
@@ -2031,7 +2031,7 @@ def save_function(pickler, obj):
             fattr = getattr(obj, fattrname, None)
             if fattr is not None:
                 state_dict[fattrname] = fattr
-        if obj.__qualname__ != obj.__name__:
+        if obj.__qualname__ != obj.__code__.co_qualname:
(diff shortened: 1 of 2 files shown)


mmckerns:
The unfortunate thing about changing behavior is that we also have to make sure the changes to the code do not impact the ability to unpickle old pickles that were created before the change.  I'll review your changes -- can you check your changes versus some old pickles?  If it fails, we would need to add a shim to the code to handle old pickles.

mmckerns:
@dfremont: so `save_function` is decorated by `register`, so it's the function that is called when the unpickler encounters a `FunctionType`.  So, it's definitely called on unpickling.  The question is, by changing the behavior, are there now existing objects that cannot be unpickled because `__name__` != `co_qualname` / `co_name`.  I'd assume yes.  Would adding an "or" to the condition solve it, so it's still backward compatible and catches the behavior that fixes #602?

PR #759, opened 2026-06-15
open session files with O_NOFOLLOW to close symlink toctou

alhudz:
Repro: in a world-writable dir, plant a regular `session.pkl`, then race a symlink to a victim file into place in the window between `_check_symlink()`s `lstat` and the following `open()`; `load_module`/`load_module_asdict` read through the link and the `dump_module` truncating fallback writes through it.
Cause: the symlink guard and the open are two separate syscalls, so the `lstat` result is stale by the time the path is opened (a classic TOCTOU).
Fix: open with `O_NOFOLLOW` so the kernel refuses a final symlink as part of the open itself; `_check_symlink` stays as an early, friendly error. `O_NOFOLLOW` is gated through `getattr(os, ...)` so non-POSIX platforms degrade to the prior behaviour.

The added `test_symlink_toctou` swaps the file for a symlink exactly at the `lstat` instant: it follows the link (and trips on the foreign contents) before this change and raises `OSError` after.

### DIFF
--- a/dill/session.py
+++ b/dill/session.py
@@ -39,6 +39,9 @@ from ._dill import (
     _reverse_typemap, __builtin__, UnpicklingError,
 )
 
+# O_NOFOLLOW is POSIX-only; degrade to a no-op flag elsewhere (e.g. Windows).
+_O_NOFOLLOW = getattr(os, 'O_NOFOLLOW', 0)
+
 def _check_symlink(path):
     """Raise OSError if *path* is a symlink."""
     try:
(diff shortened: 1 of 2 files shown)


dill/session.py
```diff
@@ -39,6 +39,9 @@ class SecurityWarning(UserWarning):
     _reverse_typemap, __builtin__, UnpicklingError,
 )
 
+# O_NOFOLLOW is POSIX-only; degrade to a no-op flag elsewhere (e.g. Windows).
+_O_NOFOLLOW = getattr(os, 'O_NOFOLLOW', 0)
```
mmckerns, line 45: should this be: `_O_NOFOLLOW = getattr(os, 'O_NOFOLLOW', os.O_RDONLY)`?  `0 `is not guaranteed to be `os.O_RDONLY`.

alhudz: `0` is intentional here. `_O_NOFOLLOW` is only ever OR'd into the flags, and each `os.open` already sets its own access mode explicitly (`O_RDONLY` for the read, `O_WRONLY` for the two writes), so this constant just needs to be the OR identity. Falling back to `os.O_RDONLY` would OR a read bit into the two write/create opens, which isn't what we want. And as you say `O_RDONLY` isn't guaranteed to be `0`, which is the reason I kept the neutral `0` rather than an access mode. I've pushed a comment spelling that out.

PR #762, opened 2026-07-09
look up descriptor by objclass in _getattr instead of eval

alhudz:
`_getattr` reconstructs slot/getset/method descriptors on unpickle.

Repro: `_getattr(object, '__doc__"].count("zz") or setattr(__import__("dill"),"_pwned",True) or object.__dict__["__doc__', "x'x'x'object'x")` runs the injected `setattr`; a crafted pickle reducing to `_getattr` reaches the same eval on `load`.
Cause: the fast path `eval(attr+'.__dict__["'+name+'"]')` evaluates a string built from the pickled `repr_str`/`name`; the `except` branch already returns the identical descriptor via `objclass.__dict__[name]`.
Fix: drop the `eval` branch and always resolve through `objclass`. `repr_str` is now unused but kept in the signature so existing pickles still load.

### DIFF
--- a/dill/_dill.py
+++ b/dill/_dill.py
@@ -1032,23 +1032,20 @@ def _create_capsule(pointer, name, context, destructor):
         return capsule
 
 def _getattr(objclass, name, repr_str):
-    # hack to grab the reference directly
-    try: #XXX: works only for __builtin__ ?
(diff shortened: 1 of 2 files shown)


alhudz:
any update?

mmckerns:
On my todo list.

PR #771, opened 2026-07-26
ensure valid names before the verification exec in _getimport

alhudz:
Repro: `obj.__module__ = 'os'; obj.__name__ = "path\nimport dill; dill._injected = True\n#"`, then `getimport(obj)` runs the injected line; `importable`, `getimportable` and `likely_import` reach the same spot.
Cause: `_getimport` interpolates `head`/`tail` into `from head import tail` and `exec`s it to verify the import, so a `__name__` carrying a newline is executed instead of rejected.
Fix: require both to be dotted identifiers before that `exec` and raise `SyntaxError`, as `getsource` already does for a bad qualname in d948ecd. `verify=False` never execs, so that path is untouched.


alhudz:
any update on this?

PR #772, opened 2026-07-31
escape names interpolated into source-matching patterns

alhudz:
Repro: set `Baz.__name__ = r'\w+'` on any class and call `dill.source.getsource(Baz)`. It returns the source of a different class defined later in the same file. With `Baz.__name__ = 'Baz(('` it raises `re.error` rather than the `IOError` the function documents.

Cause: `findsource` concatenates `__name__` straight into the class-definition pattern, so a name that is not an identifier is matched as a pattern instead of literally. `__name__` is freely assignable and is not required to be an identifier. `_closuredimport` builds its two candidate patterns from `getname()` the same way, and `getname()` can return a `repr` containing brackets even for ordinary objects.

Fix: `re.escape` the interpolated names at the three sites. `re.escape` leaves identifiers untouched, so behaviour for ordinary names is unchanged. Test added covers both the wrong-match and the uncompilable name.

### DIFF
--- a/dill/source.py
+++ b/dill/source.py
@@ -235,7 +235,9 @@ def findsource(object):
     except AttributeError: pass
     if isclass(object):
         name = object.__name__
-        pat = re.compile(r'^(\s*)class\s*' + name + r'\b')
+        # __name__ is not required to be an identifier, so escape it to keep
+        # it a literal rather than letting it act as part of the pattern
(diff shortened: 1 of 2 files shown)


alhudz:
pushed a test exercising the closured import path, which covers the escaped lines in `_closuredimport` that `codecov/patch` was flagging

alhudz:
any update?

PR #774, opened 2026-08-17
Revert "look up descriptor by objclass in _getattr instead of eval"

mmckerns:
Reverts uqfoundation/dill#762
(pypy test failures)


mmckerns:
@alhudz: I'm not sure why this wasn't apparent from the CI, but `pypy` fails for #762.  In my testing, I did some testing with python but not pypy. Note: right now pypy broke their serialization for v7.3.22 and v7.3.23, but is fixed in v7.3.24 (not yet released), so you'd have to test with versions prior to v7.3.22.

